PR #1, opened 2016-01-15
Add new command for checking external symbols

rmcgibbo:
Implements https://github.com/rmcgibbo/deloc8/issues/1. Here's the help text. Any corrections or rephrasings would be greatly appreciated:

```
$ auditwheel symbols -h
usage: auditwheel symbols [-h] [-w WHITELIST_FILE] WHEEL_FILE

The runtime dynamic linker that loads your Python extension module
when it's imported doesn't have the same concept of namespaces that
exists in Python, so it's possible for symbols defined in one module
to automatically, at runtime, override symbols defined or used in
another file, in ways that the author of that module probably didn't
intend.

(This why tricks like LD_PRELOAD work.)

It's therefore good practice to limit the number of dynamic symbols
that your module adds to the global symbol namespace.

This tool will print all of the (unnecessary) symbols that a wheel
exports. We recommend that you try to keep this list as short as
possible.

For more information on how to control visibility during compilation,
see https://gcc.gnu.org/wiki/Visibility. We recommend compiling with
-fvisibility=hidden and explicitly labeling only functions you want to
export with ``__attribute__ ((visibility ("default")))``.

For more information on the meaning of STV_DEFAULT, STB_GLOBAL, and
the like, see
http://www.sco.com/developers/gabi/2000-07-17/ch4.symtab.html.

positional arguments:
  WHEEL_FILE            Path to wheel file.

optional arguments:
  -h, --help            show this help message and exit
  -w WHITELIST_FILE, --whitelist WHITELIST_FILE
                        Skip any symbol names that appear in this file
                        (plaintext, one per line).
```

Also, @njsmith, here's the output for numpy: https://gist.github.com/rmcgibbo/4b06409dd7f6d16a8bf2

### DIFF
--- a/auditwheel/main.py
+++ b/auditwheel/main.py
@@ -7,7 +7,7 @@ import pkg_resources
 from . import main_show
 from . import main_addtag
 from . import main_lddtree
-
+from . import main_symbols
 
 def main():
     dist = pkg_resources.get_distribution('auditwheel')
@@ -28,6 +28,7 @@ def main():
     main_show.configure_parser(sub_parsers)
     main_addtag.configure_parser(sub_parsers)
     main_lddtree.configure_subparser(sub_parsers)
+    main_symbols.configure_subparser(sub_parsers)
 
     args = p.parse_args()
 
--- a/auditwheel/main_symbols.py
+++ b/auditwheel/main_symbols.py
@@ -1,4 +1,4 @@
-"""The runtime dynamic linker that loads your Python extension module when it's imported doesn't have the same concept of namespaces that exists in Python, so it's possibe for symbols defined in one module to automatically, at runtime override symbols defined or used in another file, in ways that the author of that module probably didn't intend.
+"""The runtime dynamic linker that loads your Python extension module when it's imported doesn't have the same concept of namespaces that exists in Python, so it's possibe for symbols defined in one module to automatically, at runtime, override symbols defined or used in another file, in ways that the author of that module probably didn't intend.
 
 (This why tricks line LD_PRELOAD work.)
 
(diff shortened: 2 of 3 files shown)


rmcgibbo:
Here's a nice hack:

```
import os
import ctypes

with open('hack.c', 'w') as f:
    print('''
#include "stdio.h"
#include "stdlib.h"
void npy_sin() {
    printf("hahaha crash!\\n");
    exit(1);
    }''', file=f)

os.system('gcc -shared -o libhack.so -fPIC hack.c')
ctypes.CDLL('./libhack.so', mode=ctypes.RTLD_GLOBAL)

import numpy as np
np.sin(np.pi)
```

```
$ python evil.py
hahaha crash!
````

**Comment by rmcgibbo** [comment]
But I wasn't able to get this to kill to show any bad effects:

```
$ cat spam.c
#include "stdio.h"
#include "stdlib.h"
#include <Python.h>
void npy_sin() {
    printf("hahaha crash!\n");
    exit(1);
}
static struct PyModuleDef spammodule = {
    PyModuleDef_HEAD_INIT,
    "spam", NULL, -1, NULL
};
PyMODINIT_FUNC
PyInit_spam(void) {
    return PyModule_Create(&spammodule);
}

rmcgibbo@vspm9  ~/projects/test-linkerstuff (3.4.2)
$ gcc -shared -o spam.so -fPIC spam.c -I$HOME/miniconda/include/python3.4m

rmcgibbo@vspm9  ~/projects/test-linkerstuff (3.4.2)
$ python -c 'import spam; import numpy; print(numpy.sin(1))'
0.841470984808
```

so now I'm actually not sure if this public symbol stuff is such a big deal.

njsmith:
See the email I just sent to manylinux-discuss...

Issue #8, opened 2016-01-30
Check UCS2 vs UCS4


rmcgibbo:
For Py < 3.2, the manylinux1 policy requires building against UCS4 Python, so we have to check for this in extensions modules by looking through the symbol table for the use of any `PyUnicodeUCS2_*` functions, and give a helpful error message.

njsmith:
I just built UCS2 and UCS4 versions of 2.7 to check, and can confirm that yeah, the only differences in the exported symbols are that UCS2 builds export a bunch of `PyUnicodeUCS2_*` and `_PyUnicodeUCS2_*` functions, and UCS4 builds do not.

rmcgibbo:
Well, the other difference is that the UCS4 builds export `PyUnicodeUSC4_*` symbols. But yeah, to implement this, I'm going to do the Python equivalent of

`readelf --dyn-syms <extension.so> | grep PyUnicodeUCS4 | wc`

njsmith:
Oh, sure, but "not using UCS4" is fine, it's "using UCS2" that's bad :-)

njsmith:
(Speaking of which I assume in your comment that's a typo and you mean UCS2 in the grep command.)

rmcgibbo:
Yeah, I edited it right after posting it.

njsmith:
Ah-hah, the edit just didn't propagate to my screen. It's very confusing how some parts of github pages automatically update as things change, and some parts just go stale :-(

PR #9, opened 2016-01-31
Implement UCS2 checks

rmcgibbo:


rmcgibbo:
@njsmith, I'm just going to add `--no-check-certificate`, but this may or may not be a misconfiguration of your server:

```
Connecting to vorpus.org (vorpus.org)|69.93.127.174|:443... connected.
ERROR: cannot verify vorpus.org's certificate, issued by `/C=XX/ST=There is no such thing outside US/L=Everywhere/O=OCOSA/OU=Office for Complication of Otherwise Simple Affairs/CN=roberts.vorpus.org/emailAddress=root@roberts.vorpus.org':
  Self-signed certificate encountered.
    ERROR: certificate common name `roberts.vorpus.org' doesn't match requested host name `vorpus.org'.
To connect to vorpus.org insecurely, use `--no-check-certificate'.
```

njsmith:
Yeah, misconfiguration, sorry -- use `https://vorpus.org` instead of `https://roberts.vorpus.org`

rmcgibbo:
The code is using `https://vorpus.org` https://github.com/manylinux/auditwheel/blob/usc2/tests/test-wheelhouses.sh#L11-L12

auditwheel/policy/external_references.py
```diff
@@ -13,8 +13,8 @@
 LIBPYTHON_RE = re.compile('^libpython\d\.\dm?.so(.\d)*$')
 
 
-def elf_exteral_referenences(elftree: Dict, wheel_path: str):
-    # XXX: Document the elftree structure, or put it in something
+def lddtree_exteral_referenences(lddtree: Dict, wheel_path: str):
```
njsmith: FYI `external` and `references` both appear to be misspelled :-)

njsmith:
well, wtf, then, that's weird -- are you somehow using a http client that doesn't do SNI? because you seem to be getting the self-signed certificate that debian autogenerates for unrecognized vhosts, and if you visit that URL in a web browser then you definitely get a real, not-self-signed certificate valid for vorpus.org (and with no mention of roberts.vorpus.org).

...Though, uh. Stepping back, I'm not sure I'm comfortable with my little VPS there becoming critical infrastructure :-)

For tests, maybe it would better to put together a small collection of small modules that break the rules in known, different ways? E.g. a little Cython module built (a) correctly, (b) linking to an illegal library, (c) with bad symbol versions, (d) with a UCS-2 build, and then just check the resulting wheels into the tree?

rmcgibbo:
> ...Though, uh. Stepping back, I'm not sure I'm comfortable with my little VPS there becoming critical infrastructure :-)

Yes, that's fair. The tests I have were kind of just throw together randomly. I did add a wheel satisfying property (d) to the tree in this PR though.

rmcgibbo:
But there are lots of possible corner cases, so there is some virtue in the shotgun approach for testing.

rmcgibbo:
Closes #8

PR #17, opened 2016-02-09
WIP: Update to latest PEP draft

rmcgibbo:
Currently chages `auditwheel show` output to something like

```
$ auditwheel show tests/cffi-1.5.0-cp27-none-linux_x86_64.whl
Symbol Versioning
-----------------
The wheel references the following external versioned symbols in
system-provided shared libraries: GLIBC_2.3.
This is consistent with the manylinux1 tag.

External Libraries
------------------
The following external shared libraries are required by the wheel:
{
    "libc.so.6": "/lib/x86_64-linux-gnu/libc-2.21.so",
    "libffi.so.5": null,
    "libpthread.so.0": "/lib/x86_64-linux-gnu/libpthread-2.21.so"
}
This is inconsistent with the manylinux1 tag.
In order to achieve the tag platform tag "manylinux1_x86_64" the following
shared library dependencies will need to be eliminated:
libffi.so.5

Unicode
-------
This wheel has a "none" ABI tag. Rebuild with the latest
release of ``pip install wheel`` to fix. This is inconsistent
with the manylinux1 tag.
```

### DIFF
--- a/auditwheel/elfutils.py
+++ b/auditwheel/elfutils.py
@@ -37,18 +37,6 @@ def elf_find_versioned_symbols(elf: ELFFile) -> Iterator[Tuple[str, str]]:
                        vernaux.name.decode('utf-8'))
 
 
-def elf_find_ucs2_symbols(elf: ELFFile) -> Iterator[str]:
-    section = elf.get_section_by_name(b'.dynsym')
-    if section is not None:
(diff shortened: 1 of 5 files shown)


ehashman:
I'm gonna close this because it's stale and it's unmergeable. I imagine this is superseded by manylinux2010?

PR #20, opened 2016-03-16
fix max versioned symbol

pvanderlinden:
The method `max_versioned_symbol` below crashes when the symbol has an underscore in the name, for example `HDF5_SERIAL_1.8.13`.


rmcgibbo:
Thanks!

Issue #24, opened 2016-03-29
rename vendored libraries


njsmith:
as per here:
https://mail.python.org/pipermail/wheel-builders/2016-March/000082.html

njsmith:
On further investigation, it looks like there are actually five required steps to properly vendor an ELF library:

1) Move vendored library into `.libs`
2) Run `patchelf --set-rpath path/to/.libs OBJ` on all the objects that link to the vendored library
3) Rename the library, e.g. from `libfoo.so.1` to `libfoo-<unique>.so.1`
4) Run `patchelf --replace-needed libfoo.so.1 libfoo-<unique>.so.1 OBJ` on all the objects that link to the vendored library, so that they look for the new name.
5) Crucial, but not at all obvious: Run `patchelf --set-soname libfoo-<unique>.so.1 .libs/libfoo-<unique>.so.1` so that the renamed library knows what its new name is. (The DT_SONAME attribute is what's checked when deciding whether to short-circuit later library lookups, so this is crucial -- if we didn't change this, then when someone else came along looking for `libfoo.so.1` then it would think that the already-loaded `libfoo-<unique>.so.1` satisfied that.)

njsmith:
Also, NB that `--replace-needed` and `--set-soname` are new features in `patchelf` 0.9, which is not yet shipped by e.g. Debian.

njsmith:
@rmcgibbo @ogrisel: looks like this is now the one big blocker for mass uploading of manylinux wheels now... any chance one of you has a chance to look at it?

rmcgibbo:
Yeah, I think I can handle it tonight or tomorrow day.

matthew-brett:
Great - that would be a big help.

PR #26, opened 2016-04-01
Integration test for auditwheel repair

ogrisel:
This uses 2 docker containers, one to build numpy and repair the wheel to make it manylinux1 compatible and the other to quickly check that numpy can be installed and works.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ python:
   - "3.5"
 
 script:
-  - py.test
+  - py.test -s
   - bash tests/test-wheelhouses.sh
   - auditwheel lddtree $(python -c 'import sys; print(sys.executable)')
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 language: python
+sudo: required
+services:
+  - docker
 
 notifications:
   email: false
--- a/tests/test_manylinux.py
+++ b/tests/test_manylinux.py
@@ -1,0 +1,142 @@
+from subprocess import check_call, check_output, CalledProcessError
+import pytest
+import os
+import os.path as op
+import tempfile
+import shutil
+import warnings
+from distutils.spawn import find_executable
+
+VERBOSE = True
+ENCODING = 'utf-8'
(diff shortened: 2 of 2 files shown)


ogrisel:
Travis is green \o/. Note that I chose not to run the full numpy test suite but just run a simple program that does GEMM (BLAS) and SVD (LAPACK).

njsmith:
This is pretty awesome!

It would be great if you could add a regression test specifically for the gfortran problem, though, since patchelf is a bit of a hairy program, and they just merged my patch without any tests :-). (In fact they have no tests at all for `--replace-needed`.) It seems like it shouldn't be that hard -- you could just run the numpy f2py tests if you're feeling lazy? (It might even be worth running the full numpy test suite -- I'll let Robert make the judgement call there :-).)

I guess there's also a question of whether we want to test this here or in the manylinux docker build itself. Or both...

ogrisel:
We would need to build the master branch of patchelf in the manylinux docker image but this is a pain to build as the autotools of centos:5.11 are too old to run the bootstrap script. We could host our own bootstrapped tarball in the meantime though.

ogrisel:
Alternatively it would be possible to do a bit of sed woodo in the Dockerfile to remove the unsupported autotools construct from the master clone of patchelf.

Anyway I think we can already merge this PR as it is and rebase #24 on top of it to add the patchelf tests.

matthew-brett:
I built a tarball with Nathaniel's patches for the tests last night : http://nipy.bic.berkeley.edu/manylinux/patchelf-0.9njs.tar.gz

rmcgibbo:
Do we want to build a proper python / pip-installable interface to patchelf? It would be straightforward to either provide a python library / API that internally invoked the binary or perhaps a little trickier but still straightforward to modify the code to provide an API that we wrap from cython.

ogrisel:
> Do we want to build a proper python / pip-installable interface to patchelf? It would be straightforward to either provide a python library / API that internally invoked the binary or perhaps a little trickier but still straightforward to modify the code to provide an API that we wrap from cython.

I am not sure it's worth writing a Python library on top of it. The command line interface seems good enough to me.

PR #27, opened 2016-04-14
fix invalid manylinux wheel filename for 32 bits platform

anlambert:
When repairing Python wheel on 32 bits linux, the resulting wheel file ends suffixed in a wrong way : 'manylinux1__i686' instead of 'manylinux1_i686'. The problem comes from the get_arch_name in the policy sub module, looks like a typo to me.


ogrisel:
Thank you very much. Indeed this is a typo I made during the last refactoring. I should have tested it better.

PR #33, opened 2016-04-21
FIX make sure to run the tests with the latest docker images

ogrisel:
I got bitten running the tests on an old manylinux1 cached docker image on my local machine. I don't think travis is impacted.


ogrisel:
@rmcgibbo @njsmith merge?

Issue #32, opened 2016-04-15
check for binaries in purelib


njsmith:
To help in situations like #28, it would be nice if auditwheel could check for impure files in purelib and complain if they're found.

PR #35, opened 2016-04-22
check for invalid binary wheel

anlambert:
That PR is there to adress issue #32. No binaries should be found in the purelib folder of a wheel. So raise an error if it is the case.

### DIFF
--- a/auditwheel/wheel_abi.py
+++ b/auditwheel/wheel_abi.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import functools
+import os
 from os.path import basename
 from typing import Dict, Set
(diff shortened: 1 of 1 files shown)


ogrisel:
LGTM. @njsmith @rmcgibbo what do you think?

Shall we try to write an integration test to check the behavior of auditwheel for such invalid inputs? That sounds complex to do.

njsmith:
A test would indeed be good... it should just be a matter of creating a tiny broken wheel (e.g. using @anlambert's technique of packing a .so into a wheel file as if it were just a data file) and saving it in the repo to run the test agains?

PR #38, opened 2016-05-10
Fix manylinux #60 bug

rmcgibbo:
Fixes https://github.com/pypa/manylinux/issues/60

### DIFF
--- a/auditwheel/wheel_abi.py
+++ b/auditwheel/wheel_abi.py
@@ -40,8 +40,9 @@ def get_wheel_elfdata(wheel_fn: str):
                 if py_ver == 2:
                     uses_ucs2_symbols |= any(
                         True for _ in elf_find_ucs2_symbols(elf))
-                full_external_refs[fn] = lddtree_external_references(elftree,
-                                                                     ctx.path)
+
+            full_external_refs[fn] = lddtree_external_references(elftree,
+                                                                 ctx.path)
 
     log.debug(json.dumps(full_elftree, indent=4))
     return (full_elftree, full_external_refs,


rmcgibbo:
If anyone could figure out how to add a test for https://github.com/pypa/manylinux/issues/60 to the travis test suite here, that would be excellent. I could copy the un-repaired wheel into the test data suite, but we'd still have to download the system-libraries-to-be-grafted-in as well.

ogrisel:
travis failed: elftree is undefined in some cases.

ogrisel:
For the integration tests, I think we should write a bash script that would download and install libhfd5 from the source tarball into the manylinux1 container.

njsmith:
> If anyone could figure out how to add a test

I don't have the time today to grok what the underlying issue actually is, but I've had good luck in the past with writing my own tiny tiny .so's to exercise weird linker issues -- like you can generate a set of .so's with a funky transitive dependency with just this much writing:
```
-- a.c --
extern int from_b();
extern int from_c();
int from_a() {
    return from_b() + from_c();
}

-- b.c --
extern int from_c();
int from_b() {
    return 2 * from_c();
}

-- c.c --
int from_c() {
    return 1;
}

-- build.sh --
gcc c.c -o c.so -shared
gcc b.c -o b.so -shared -L. -lc
# a.so links directly to b.so but not to c.so
gcc a.c -o a.so -shared -L. -lb
```

ehashman:
I *think* I included this fix in #110 but with a little more rigour and some additional tests so I'm going to close this.

PR #45, opened 2016-06-06
Remove __init__.py check

soltanmm:
Maybe needs a test for namespace packages?

Not necessarily volunteering - if it isn't immediately necessary I'd rather just continue on my merry way. :stuck_out_tongue_closed_eyes:


njsmith:
The change makes sense to me but a test would be good, yes :-)

PR #55, opened 2016-11-17
Spelling

tdsmith:
Just happened to notice.


matthew-brett:
Thanks for taking the time to fix that.

Issue #56, opened 2016-12-24
Auditwheel should check for references to PyFPE_jbuf, and error out if present


njsmith:
See:
https://github.com/python/peps/pull/166
https://github.com/numpy/numpy/issues/8415#issuecomment-269103493

rmcgibbo:
Fixed.

PR #58, opened 2016-12-25
Check for PyFPE_jbuf and error out if present

rmcgibbo:
Potential fix for #56.

Unfortunately, I haven't yet tested this code against a wheel that _does_ contain a reference to `PyFPE_jbuf`. Could anyone potentially provide a link to such a wheel so that we can include it as a test?


njsmith:
Here's a quick one (wrapped in a .zip to make github happy): [dist.zip](https://github.com/pypa/auditwheel/files/672077/dist.zip)

From https://github.com/njsmith/fpewheel, then built with Debian's python 3.5.2-9.

auditwheel/elfutils.py
```diff
@@ -64,6 +64,18 @@ def elf_find_ucs2_symbols(elf: ELFFile) -> Iterator[str]:
                 yield sym.name
 
 
+def elf_references_PyFPE_jbuf(elf: ELFFile) -> bool:
+    section = elf.get_section_by_name('.dynsym')
+    if section is not None:
+        # look for UCS2 symbols that are externally referenced
```
njsmith: comment's out of date

njsmith:
I guess technically maybe we should also check for references to `PyFPE_dummy` and `PyFPE_counter`, but in practice either they're all referenced or none of them are so it doesn't matter much.

njsmith:
Is this ready to merge, or do you want to add a test first?

rmcgibbo:
I've just been on holiday, but I'll add a test when I get back (next couple days).

tests/test_pyfpe.py
```diff
@@ -0,0 +1,8 @@
+from auditwheel.wheel_abi import analyze_wheel_abi
+from auditwheel.policy import POLICY_PRIORITY_LOWEST
+
+
+def test_analyze_wheel_abi():
+    winfo = analyze_wheel_abi('tests/fpewheel-0.0.0-cp35-cp35m-linux_x86_64.whl')
+    assert winfo.sym_tag == 'manylinux1_x86_64'  # for extrernal symbols, it could get manylinux1
```
ogrisel: typo: "external"

PR #70, opened 2017-05-04
Fix exception raised for certain symbols

szilveszter:
I was trying to repair wheels targeted for linux (not manylinux) on a more recent distro, and couldn’t get the cryptography library repaired, because it contained versioned symbols the current code could not handle:

```python
defaultdict(<function get_wheel_elfdata.<locals>.<lambda> at 0x7f90deb869d8>, {'libc.so.6': {'GLIBC_2.2.5', 'GLIBC_2.14', 'GLIBC_2.4'}, 'libcrypto.so.10': {'libcrypto.so.10', 'OPENSSL_1.0.1_EC', 'OPENSSL_1.0.1'}, 'libssl.so.10': {'libssl.so.10'}})
```

This change makes sure that these symbols can be successfully processed.


szilveszter:
Nevermind, this seems to result in unusable wheel files.

PR #73, opened 2017-05-23
support versioned symbols with more underscores

### LINKED ISSUE #72: ValueError in versioned symbols
When compiling a manylinux1 wheel against OpenSSL 1.1.0 the following exception is raised:

```
  File "/usr/local/bin/auditwheel", line 11, in <module>
    sys.exit(main())
  File "/opt/_internal/cpython-3.6.0/lib/python3.6/site-packages/auditwheel/main.py", line 49, in main
    rval = args.func(args, p)
  File "/opt/_internal/cpython-3.6.0/lib/python3.6/site-packages/auditwheel/main_repair.py", line 60, in execute
    wheel_abi = analyze_wheel_abi(args.WHEEL_FILE)
  File "/opt/_internal/cpython-3.6.0/lib/python3.6/site-packages/auditwheel/wheel_abi.py", line 82, in analyze_wheel_abi
    symbol_policy = versioned_symbols_policy(versioned_symbols)
  File "/opt/_internal/cpython-3.6.0/lib/python3.6/site-packages/auditwheel/policy/versioned_symbols.py", line 28, in versioned_symbols_policy
    sym_name, _ = symbol.split("_", 2)
ValueError: too many values to unpack (expected 2)
```

Looking in PDB the issue is that the `symbol` is expected to look something like `GLIBC_2.2.5`, but OpenSSL 1.1.0 is `OPENSSL_1_1_0`, which results in too many values.

reaperhulk:
Fixes #72. I'm unsure if this test is useful so let me know if I should be doing something different.

### DIFF
--- a/auditwheel/policy/versioned_symbols.py
+++ b/auditwheel/policy/versioned_symbols.py
@@ -24,7 +24,7 @@ def versioned_symbols_policy(versioned_symbols: Dict[str, Set[str]]) -> int:
     required_vers = {}  # type: Dict[str, Set[str]]
     for symbols in versioned_symbols.values():
         for symbol in symbols:
-            sym_name, _ = symbol.split("_", 2)
+            sym_name, _ = symbol.split("_", 1)
             required_vers.setdefault(sym_name, set()).add(symbol)
     matching_policies = []  # type: List[int]
     for p in load_policies():
--- a/auditwheel/policy/versioned_symbols.py
+++ b/auditwheel/policy/versioned_symbols.py
@@ -24,7 +24,7 @@ def versioned_symbols_policy(versioned_symbols: Dict[str, Set[str]]) -> int:
     required_vers = {}  # type: Dict[str, Set[str]]
     for symbols in versioned_symbols.values():
         for symbol in symbols:
-            sym_name, _ = symbol.split("_", 1)
+            sym_name, _, _ = symbol.partition("_")
             required_vers.setdefault(sym_name, set()).add(symbol)
     matching_policies = []  # type: List[int]
     for p in load_policies():
--- a/tests/test_policy.py
+++ b/tests/test_policy.py
@@ -18,3 +18,5 @@ def test_policy_checks_glibc():
     assert policy == POLICY_PRIORITY_LOWEST
     policy = versioned_symbols_policy({"some_library.so": {"OPENSSL_1_1_0"}})
     assert policy == POLICY_PRIORITY_HIGHEST
+    policy = versioned_symbols_policy({"some_library.so": {"IAMALIBRARY"}})
+    assert policy == POLICY_PRIORITY_HIGHEST
--- a/tests/test_policy.py
+++ b/tests/test_policy.py
@@ -1,6 +1,8 @@
 from jsonschema import validate
 from auditwheel.policy import (load_policies, _load_policy_schema,
-                               versioned_symbols_policy, POLICY_PRIORITY_LOWEST)
+                               versioned_symbols_policy,
+                               POLICY_PRIORITY_HIGHEST,
+                               POLICY_PRIORITY_LOWEST)
 
 
 def test_policy():
@@ -14,3 +16,5 @@ def test_policy_checks_glibc():
     assert policy > POLICY_PRIORITY_LOWEST
     policy = versioned_symbols_policy({"some_library.so": {"GLIBC_999"}})
     assert policy == POLICY_PRIORITY_LOWEST
+    policy = versioned_symbols_policy({"some_library.so": {"OPENSSL_1_1_0"}})
+    assert policy == POLICY_PRIORITY_HIGHEST
--- a/auditwheel/policy/__init__.py
+++ b/auditwheel/policy/__init__.py
@@ -22,7 +22,7 @@ _PLATFORM_REPLACEMENT_MAP = {
 # XXX: this could be weakened. The show command _could_ run on OS X or
 # Windows probably, but there's not much reason to inspect foreign package
 # that won't run on the platform.
-if platform == 'linux':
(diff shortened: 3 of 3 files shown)


tests/test_policy.py
```diff
@@ -14,3 +16,5 @@ def test_policy_checks_glibc():
     assert policy > POLICY_PRIORITY_LOWEST
     policy = versioned_symbols_policy({"some_library.so": {"GLIBC_999"}})
     assert policy == POLICY_PRIORITY_LOWEST
+    policy = versioned_symbols_policy({"some_library.so": {"OPENSSL_1_1_0"}})
+    assert policy == POLICY_PRIORITY_HIGHEST
```
rmcgibbo: Since `OPENSSL_1_1_01 isn't in the whitelist, I might have expect this to give POLICY_PRIORITY_LOWEST, but perhaps there's something I'm not thinking of.

rmcgibbo:
Does auditwheel with this patch repair your openssl-linked wheel properly?

tests/test_policy.py
reaperhulk: Yeah I don't understand the policy priorities so I have no idea if this assertion is correct!

reaperhulk:
`auditwheel` does properly repair my openssl-linked wheel with this patch applied.

auditwheel/policy/versioned_symbols.py
```diff
@@ -24,7 +24,7 @@ def policy_is_satisfied(policy_name: str,
     required_vers = {}  # type: Dict[str, Set[str]]
     for symbols in versioned_symbols.values():
         for symbol in symbols:
-            sym_name, _ = symbol.split("_", 2)
+            sym_name, _ = symbol.split("_", 1)
```
ehashman: Yay fixing off by 1 error.

tests/test_policy.py
ehashman: @rmcgibbo: So the policy whitelist only applies to GCC*/CXX* symbol versions (see `auditwheel/policy/policy.json` for the full list). Any other symbol version will automatically satisfy the highest policy. If that wasn't the case, then no external dependency symbols (e.g. `OPENSSL_1_1_0`, `KERBEROS_5_MIT`) would satisfy the manylinux1 requirements.

ehashman: Would you mind adding one more test asserting `SYMBOLWITHOUTUNDERSCORES` satisfies `POLICY_PRIORITY_HIGHEST` or something similar?

reaperhulk: This will in fact fail since the assignment assumes 2 values. Is it possible for a symbol to be entirely unversioned? I guess the answer to that is likely to be yes?

reaperhulk: We can use `partition` here to do this. I'll swap it.

ehashman:
@rmcgibbo Do you mind cutting a release for this? I would if I had access :)

matthew-brett:
@ehashman - I guess you mean pypi access?  Sounds like a good idea to me.  Thanks for all the work so far.

ehashman:
Yeah, tagging and uploading a release to PyPI. I don't know what the process is so I figured I'd ask.

matthew-brett:
I think you'll need @rmcgibbo or @ogrisel to give you maintainer permissions on pypi, in order to upload.

@rmcgibbo - would you mind throwing a few notes about release procedure down, either here or in a text file somewhere in the distribution?

rmcgibbo:
@ehashman: what's you username on pypi.org?

ehashman:
I am also ehashman there.

rmcgibbo:
The release procedure I've used so far is pretty simple.
1. The project's setup.py/setup.cfg uses `pbr`, so to increment the version number, you just make a new git tag. So something like `git tag -s 1.6 -m 'Release version 1.6' I think, and `git push origin --tags`.
2. And then I've been uploading sdists and wheels to PyPI (`python setup.py sdist upload` && `python setup.py bdist_wheel upload`).

rmcgibbo:
@ehashman: great. you should be listed as an auditwheel maintainer on pypi.org now.

ehashman:
Released.

matthew-brett:
Excellent - thanks.

njsmith:
And I just submitted the PR to rev the manylinux docker images: https://github.com/pypa/manylinux/pull/116

@ehashman: just so you know what's going on here – the manylinux docker images are always install the latest auditwheel from PyPI when they're re-built, but they only get re-built when a commit lands in the pypa/manylinux master branch (which happens via a magic travis job that pushes the new images to quay.io). Generally we follow the one-person-submits-PR, different-person-hits-merge model. As a member of the "manylinux team" you should have write access to the manylinux repo, so you could hit merge, or in the future if you do another auditwheel release it might make sense to submit a PR like this as part of the release checklist.

It suddenly occurs to me that we might be able to trick requires.io into submitting these PRs for us, even though it doesn't really fit their normal model...

reaperhulk:
@njsmith Travis has a beta feature where you can set up scheduled builds now too, so it might make sense just to have it rebuild and push to quay every day? If nothing has changed then all the layers will be identical and it will essentially be a noop.

njsmith:
@reaperhulk: Interesting point. I suppose a 0-24 hour delay in getting new auditwheels into the docker image wouldn't be a big deal... in theory it's nice to be quick, but it's not like we've been reliably quick with the current system either! Nightly builds would also start failing immediately as soon as one of our dependencies breaks (e.g. openssl releases a new version and moves the old one, or when centos5 went EOL and they moved the repo). I'm not 100% sure whether this is an advantage or disadvantage :-). I suppose this wouldn't hurt anything, just might mean annoying email for whoever is configured to get those. Maybe we wouldn't even notice at all :-).

PR #74, opened 2017-05-23
Simplify code slightly

alex:
`set(dict.keys())` is the same as `set(dict)`. This shortens the line and lets us remove linewrapping


njsmith:
Looks obviously correct to me, but I'll wait for Travis to confirm :-)

alex:
@njsmith tests confirm :-)

PR #81, opened 2017-08-25
Added libresolv to the whitelisted libraries

### LINKED ISSUE #80: Whitelist libresolv?
This library is part of glibc (similarly to libc, libm, librt, libpthread etc which are already whitelisted) and uses the same symbol versioning. I wonder if perhaps it is better not to bundle it?

Context: I ran into some trouble with the psycopg2 wheel, please see the details here: https://github.com/psycopg/psycopg2-wheels/issues/2. If I replace the bundled version with the system version it works fine.

dvarrazzo:
Fix #80

### DIFF
--- a/auditwheel/policy/policy.json
+++ b/auditwheel/policy/policy.json
@@ -24,6 +24,6 @@
          "libc.so.6", "libnsl.so.1", "libutil.so.1", "libpthread.so.0",
          "libX11.so.6", "libXext.so.6", "libXrender.so.1", "libICE.so.6",
          "libSM.so.6", "libGL.so.1", "libgobject-2.0.so.0",
-         "libgthread-2.0.so.0", "libglib-2.0.so.0"
+         "libgthread-2.0.so.0", "libglib-2.0.so.0", "libresolv.so.2"
      ]}
 ]


njsmith:
Ugh, it looks like the Travis tests have bitrotted. @rmcgibbo @ehashman any ideas?

ehashman:
Hmmmm, at first glance I don't think this is bitrotted. If we add a new file to the policy we'll need to also ensure all its symbols are represented in the policy list. Perhaps that's why this failed?

njsmith:
The library's part of glibc so it shouldn't need any new symbols.

From a quick glance it looked like the tests were failing because they were trying to fetch some remote url and getting a 404.

ehashman:
*looks at Travis build*

Aha, I dunno what https://github.com/getnikola/wheelhouse/archive/v2.7.zip is but it's 404ing. That definitely isn't a test I wrote. Will let y'all take a look first.

cblackwell-dw:
Is anyone working on fixing these build problems? They seem unrelated to this pull request. This issue affects other libraries that we are building our own wheels for; for example mysqlclient.

rmcgibbo:
I don't think anyone is working on fixing the build problems. It's my code / my fault. The 404 is surely because I found some wheels on github to use as tests.

If anyone has the bandwidth to clean up the tests a little bit -- even just remove the failing test -- that would be great.

ehashman:
I'll see if I can just remove it for now.

rmcgibbo:
(cc https://github.com/pypa/auditwheel/pull/86)

ehashman:
Hi @dvarrazzo, you can now rebase this PR and the build should pass!

dvarrazzo:
Ready to merge

PR #86, opened 2017-11-07
Remove test-wheelhouse.sh (broken tests)

rmcgibbo:


ehashman:
@rmcgibbo rather than totally removing this, do we want to modify the Travis build matrix to make it an optional test?

rmcgibbo:
It's not that the URLs are flaky. I think both pages are simply gone.

matthew-brett:
How about using the wheels at https://3f23b170c54c2533c070-1c8a9b3114517dc5fe17b7c3f8c63a43.ssl.cf2.rackcdn.com ?  Those should be around for a longish time (it's the Rackspace account belonging to scikit-learn).  Or pulling selected wheels from pypi?

ehashman:
@rmcgibbo yeah, as @matthew-brett suggested, I was thinking we could maybe update the URLs to use different wheels?

njsmith:
I was worried that that was going to go away because Rackspace was stopping doing free hosting for OSS, but apparently they backpedaled on that: https://www.theregister.co.uk/2017/10/20/rackspace_ends_discount_hosting_for_open_source_projects/

ehashman:
Opened #87 as an alternative to this PR per my earlier suggestion. @rmcgibbo let me know if you'd prefer that or this one!

PR #88, opened 2018-01-31
Manylinux2

markrwilliams:
This teaches `auditwheel` about `manylinux`. See https://github.com/pypa/manylinux/pull/152 for the corresponding Docker image.

PR #92, opened 2018-05-15
Manylinux2010

wtolson:
This is an update of @markrwilliams merge request pypa/auditwheel#88. Updates the tag from manylinux2 to manylinux2010 to reflect the changes in PEP 571. Teaches auditwheel about manylinux2010. See pypa/manylinux#152 for the corresponding Docker image.


ehashman:
@wtolson, @markrwilliams, thanks for putting this together. At a glance this all looks great to me (love the new symbol generation script), but I do want a chance to review this more closely before I hit merge. I dunno if any of the other maintainers would like to take a look.

tests/test_manylinux.py
```diff
@@ -9,7 +9,7 @@
 
 VERBOSE = True
 ENCODING = 'utf-8'
-MANYLINUX_IMAGE_ID = 'quay.io/pypa/manylinux1_x86_64'
```
mayeut: Shouldn't the tests using manylinux1 still be run ?

wtolson: I've updated the tests to run against the manlinux1 and manylinux2010 containers.

mayeut:
Tests are now showing that wheels get `manylinux2010` when built on `manylinux1` image.
Shall priority be reversed between the 2 policies ?

ofek:
Can this be merged?

ehashman:
@ofek: @njsmith pointed out to me (as does @mayeut's comment above) that manylinux2010 labels are being applied to manylinux1 compliant wheels. The order should be reversed. If a wheel passes both the manylinux1 and manylinux2010 policy, it should be labeled with the more restrictive policy (in that case, manylinux1).

So we'll need to get that fixed before this is ready to merge, otherwise this will break backwards compatibility and relabel manylinux1 wheels as manylinux2010, which we don't want.

auditwheel/policy/policy.json
```diff
@@ -25,5 +25,26 @@
          "libX11.so.6", "libXext.so.6", "libXrender.so.1", "libICE.so.6",
          "libSM.so.6", "libGL.so.1", "libgobject-2.0.so.0",
          "libgthread-2.0.so.0", "libglib-2.0.so.0", "libresolv.so.2"
+     ]},
+    {"name": "manylinux2010",
+     "priority": 200,
```
ehashman: Following up on my most recent comment, I believe this priority should be lowered such that it is between manylinux1 (most restrictive) and "linux" (least restrictive). Current manylinux1 binaries should also be manylinux2010 compliant, but not vice versa.

ehashman: Perhaps we should have done the priorities in increasing rather than decreasing order... feel free to pick an arbitrary number between 0 and 100 for now (90?) and if we need to renumber later because the range was too small we can do that eventually.

safijari:
What needs to be done to get this resolved? Is there some way I can help?

ehashman:
@safijari the policy precedence needs to be adjusted and the tests updated, per my last comment (and others)

safijari:
@ehashman Okay, I'm a little new to contributing to opensource projects, and I'm trying to see if I can help push this forward. Is it possible for me to push onto this PR? or should I open a new one?

PR #95, opened 2018-06-12
Process non-Python binaries (e.g. binary executables)

lpsinger:
This makes it possible to relocate and distribute executable binaries as well as Python extension binaries.

We would like to use this feature in the [LALSuite](https://pypi.org/project/lalsuite/) package. See https://git.ligo.org/lscsoft/lalsuite/merge_requests/164.


njsmith:
Looks plausible at a quick glance. Can you add a test though? That would make it much easier to have some confidence that it actually works :-).

PR #97, opened 2018-07-04
Add badge for PyPI releases

ehashman:


README.rst
```diff
@@ -3,6 +3,8 @@ auditwheel
 
 .. image:: https://travis-ci.org/pypa/auditwheel.svg?branch=master
     :target: https://travis-ci.org/pypa/auditwheel
+.. image:: https://badge.fury.io/py/auditwheel.svg
+    :target: https://badge.fury.io/py/auditwheel
```
njsmith: Probably should link directly to `https://pypi.io/project/auditwheel`?

ehashman: The difference is that badge.fury.io links to the exact version on the badge, but I agree with the change.

PR #104, opened 2018-10-17
Manylinux2010 Mark 2

safijari:
This PR is entirely the same as https://github.com/pypa/auditwheel/pull/92 and contains all the original commits from the original authors. All I've done is update the policy number as per the request of @ehashman. I'm not looking to step on any toes, and I certainly do not want to take credit for someone else's work, I just wanna help move this effort forward and do not know of a better way. Apologies for any (many?) mistakes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,6 @@ branches:
 
 before_install:
   - pip install --upgrade pip
-  - pip install wheel==0.31.1
 
 install:
   - pip install -r test-requirements.txt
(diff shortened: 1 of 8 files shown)


safijari:
Travis says the checks pass but I see three errors, all related to https://github.com/pypa/wheel/issues/262. The only fix for this I've seen so far is people freezing wheel to `0.31.1` or some such version.

ehashman:
The tests definitely failed, I'm not sure why the script is reporting an exit status of 0. Can you report a bug and I'll try to fix Travis later?

ehashman:
@safijari I think the tests will fail even if you get this building with the right version of wheel per #106... I would hold off on this PR until I get the regression I mentioned in #107 figured out. Working on tracking it down right now. Thanks for your help!!

safijari:
Sounds good, lemme know if I can help with anything else.

ehashman:
Hmm, looks like there are some merge conflicts now. Don't worry about this PR @safijari, I'll handle this by fixing up #92.

Issue #111, opened 2018-10-27
Drop the dot when invoking pytest


jaraco:
PyTest [recommends](https://github.com/pytest-dev/pytest/issues/1629) using `pytest` instead of `py.test`. This project should adopt that pattern in its test invocations [such as here](https://github.com/pypa/auditwheel/blob/5a56bca4f416bd069ecc03c418311c488b494f2f/tox.ini#L9).

I'm at the Bloomberg PyPA sprint and overheard 'py dot test', so thought I'd capture this trivial issue.

kaikulimu:
Hi, I will work on this.

kaikulimu:
Hi, I will work on this.

kaikulimu:
Submitted PR: https://github.com/pypa/auditwheel/pull/112

ehashman:
Closed via #112.

jaraco:
Woo hoo. Thanks!

Issue #105, opened 2018-10-18
Failing tests in Travis have incorrect exit status


safijari:
See any of the logs here https://travis-ci.org/pypa/auditwheel/builds/442957798

The build says it passed but the log shows 3 failed tests.

PR #112, opened 2018-10-27
Drop the dot when invoking pytest

kaikulimu:
In response to: https://github.com/pypa/auditwheel/issues/111


ehashman:
Thanks for this PR! I'm just testing this manually since CI is broken (see #105), and once that passes I will approve and merge.

PR #119, opened 2018-11-01
Use super() Python 3 syntax. Don't inherit from object.

atombrella:


ehashman:
Note that this code was originally copied from https://github.com/nipy/nibabel and I don't think this will break anything making these changes but at some point I'd love to go pull in all the unit tests from all the random code we vendored...

PR #123, opened 2018-11-01
Update README.rst to Python 3.4. And Python versions in setup.cfg.

atombrella:


ehashman:
At some point we should really drop 3.4 support... or at least add 3.7...

PR #127, opened 2018-11-13
Added Python 3.7 to the CI.

atombrella:


atombrella:
@ehashman It works without `sudo`, at least Travis didn't give any errors. I fixed the import-warnings of `collections` in this PR as well.

jamadden:
Using `sudo` at all [is no longer recommended by Travis](https://changelog.travis-ci.com/linux-builds-run-on-vms-by-default-77106).

ehashman:
@jamadden Previously using sudo was the only way for Docker on Travis to work. Has that changed as well?

jamadden:
Effectively, `sudo` is true by default now, and builds that *do not* specify any value for sudo run on a full VM instead of a container. [One of the originally stated reasons for that](https://blog.travis-ci.com/2018-10-04-combining-linux-infrastructures) was specifically to make using docker easier:

> More and more folks are building and deploying Docker images, and using Docker within a container-based build is not a great experience

ehashman:
There are some merge conflicts that will need to be resolved :)

ehashman:
Apparently switching the 3.5 build to xenial was the kick in the pants Travis needed. Fingers crossed this won't break as soon as it hits master again... merging.

PR #133, opened 2018-12-12
Clear rpath before setting it

### LINKED ISSUE #29: Vendored library cannot be loaded
I used the manylinux docker infrastructure to build:

http://travis-wheels.scikit-image.org/netCDF4-1.2.3.1-cp34-cp34m-manylinux1_x86_64.whl

Upon import, I see:

```
  File "<string>", line 1, in <module>
  File "/home/stefan/envs/py3/lib/python3.4/site-packages/netCDF4/__init__.py", line 3, in <module>
    from ._netCDF4 import *
ImportError: libhdf5_hl-9c7ba457.so.10.0.2: cannot open shared object file: No such file or directory
```

Looking at the vendored libraries, the RPATH is invalid for libnetcdf (not for libhdf5)--possibly has something to do with original libnetcdf library already having an RPATH entry.

Build recipe can be found here:

https://github.com/stefanv/manylinux-builds/blob/build_netcdf/build_netcdfs.sh

It was run on yesterday's version of the manylinux docker image.

/cc @njsmith @matthew-brett

daa:
Due to bug in patchelf https://github.com/NixOS/patchelf/issues/94 if shared object has ``DT_RUNPATH`` attribute it is not replaced with ``DT_RPATH`` if ``--force-rpath`` is given, so one must clear all rpath attributes before setting new. This pull request closes #29.


daa:
Strangely travis for python-3.5 failed but locally tests passed.

daa:
Also I have a question related to following code in ``auditwheel/repair.py`` where ``auditwheel`` tries to modify existing ``RUNPATH``/``RPATH``:
```
143     for rp in itertools.chain(rpaths['rpaths'], rpaths['runpaths']):
144         if is_subdir(rp, os.path.dirname(src_path)):
145             patchelf_set_rpath(dest_path, pjoin(
146                 dirname(dest_path), relpath(rp, dirname(src_path))))
147             break
```
If I understand correctly if for example ``/opt/example/libexample.so`` has ``RPATH`` set to ``/opt/example/subdir``, copied ``libexample.so`` will have ``RPATH`` set to ``$ORIGIN/subdir``, but because ``auditwheel`` does not copy libraries into subdirectories, needed libraries from ``/opt/example/subdir`` will be copied into the same destination directory as ``libexample.so`` thus making new ``RPATH`` pointing to non-existing location. Also if original ``RPATH`` is set to some another directory ``/opt/another/lib`` it will remain unmodified.
Considering said my questions are:
- What was the intention of this bit of code? And what use cases did it solve?
- Wouldn't it be better to just clear ``RPATH``/``RUNPATH`` on all copied libraries?
@ehashman @rmcgibbo

ehashman:
@daa travis has been super flaky lately for some bizarre reason, I think there's been an issue with caching. Sometimes restarting the build helps.

Thanks a bunch for submitting this (and your other recent comments!), I'll take a look as soon as I get some time next week :)

daa:
> > If I understand correctly if for example `/opt/example/libexample.so` has `RPATH` set to `/opt/example/subdir`, copied `libexample.so` will have `RPATH` set to `$ORIGIN/subdir`, but because `auditwheel` does not copy libraries into subdirectories, needed libraries from `/opt/example/subdir` will be copied into the same destination directory as `libexample.so` thus making new `RPATH` pointing to non-existing location. Also if original `RPATH` is set to some another directory `/opt/another/lib` it will remain unmodified.
> 
> I think your assessment is correct and as it is impossible to tell what the original intent of this code was, I think it's a good idea to try to simplify it, and worst case if this breaks something (unlikely given that this seems obviously buggy) we will be able to fix it and document it properly.
> 
> I do think however that this change should be in a separate PR as it is unrelated to the bug this PR is addressing. We can get it reviewed and landed quickly if you are happy to split it out.

#171

daa:
I've resubmitted this pull request as #173

PR #139, opened 2019-02-01
Recurrent dependencies

### LINKED ISSUE #136: Non-python extension dependencies with external dependencies are not recursively patched
As reported by @thomaslima:

See also #134.

thomaslima:
Closes #136 
Test cases in #134


thomaslima:
Please see my comment in https://github.com/pypa/auditwheel/pull/134#issuecomment-459899790

PR #140, opened 2019-02-01
Append to RPATH (instead of overwrite)

### LINKED ISSUE #137: RPATH is being incorrectly overwritten
As originally reported by @thomaslima:

dbcore already had an `RPATH` set to `$ORIGIN`. So because of its dependency to libcurl.so, `patchelf_set_rpath` *replaced* it to `$ORIGIN/.libs` in `repair.py`, even though it was not necessary. This way, it could no longer find `lib_tl.so`.

See also #134.

thomaslima:
Closes #137 

Test cases in #134 

Not-implemented caveat: https://github.com/pypa/auditwheel/pull/134#issuecomment-457257003

### DIFF
--- a/auditwheel/repair.py
+++ b/auditwheel/repair.py
@@ -150,6 +150,10 @@ def copylib(src_path, dest_dir):
 
 
 def patchelf_set_rpath(fn, libdir):
+    old_rpath = check_output(['patchelf', '--print-rpath', fn]
+                             ).decode('utf-8').strip()
     rpath = pjoin('$ORIGIN', relpath(libdir, dirname(fn)))
+    if old_rpath != '':
+        rpath = ':'.join([old_rpath, rpath])
     logger.debug('Setting RPATH: %s to "%s"', fn, rpath)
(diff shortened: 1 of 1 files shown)


ferdonline:
Hi @thomaslima 
Any update on this PR? In our project we also depend on internal libraries and require the previous RPATH to be respected.
Independently I arrived to a very similar solution.
```
    prev_rpath = check_output(['patchelf', '--print-rpath', fn]).decode().rstrip('\n\t /')
    check_call(['patchelf', '--remove-rpath', fn])
    if prev_rpath and prev_rpath != rpath:
        rpath += ":" + prev_rpath
```
Feel free to use it.
In this PR, however, it seems the tests fail for some unrelated reason.
We would be happy to help this getting  fixed. Thanks!

thomaslima:
@ferdonline I'd really like some help. I have absolutely no time to look into this at the moment =(

fbrennen:
Hi @thomaslima -- I had a quick go at updating your PR here: https://github.com/pypa/auditwheel/pull/245

lkollar:
#245 implemented this.

PR #147, opened 2019-04-11
Update tests for manylinux2010

takluyver:
There's now an official image and pip supports manylinux2010, so these workarounds should be unnecessary.

### DIFF
--- a/tests/test_manylinux.py
+++ b/tests/test_manylinux.py
@@ -14,8 +14,7 @@ logger = logging.getLogger(__name__)
 
 ENCODING = 'utf-8'
 MANYLINUX1_IMAGE_ID = 'quay.io/pypa/manylinux1_x86_64'
-#TODO: use the pypa manylinux2010 image on release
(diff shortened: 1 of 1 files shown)


takluyver:
Reopened because for some reason Travis wasn't noticing the second commit.

tests/test_manylinux.py
```diff
@@ -146,21 +161,15 @@ def test_build_repair_numpy(docker_container):
     docker_exec(manylinux_id, repair_command)
     filenames = os.listdir(io_folder)
 
-    if policy == 'manylinux1':
-        expected_files = 2
-    elif policy == 'manylinux2010':
-        expected_files = 3  # We end up repairing the wheel twice to manylinux1
-
-    assert len(filenames) == expected_files
-    # Regardless of build environment, wheel only needs manylinux1 symbols
```
ehashman: I was gonna ask a question about this behaviour change but it's addressed in the commit message of 46619cd0df32f9548cbdb6e1b8d6f6f9de6d07c5. Thanks for catching that.

Issue #146, opened 2019-04-11
Repair can produce an invalid manylinux1 wheel


takluyver:
I'm experimenting with building manylinux2010 wheels for pygame, to test the toolchain. `auditwheel repair` (with no `--plat`) argument tags the wheel `manylinux1`. But then re-running auditwheel on that wheel, it is not valid for manylinux1:

- `auditwheel show` says "This constrains the platform tag to "manylinux2010_x86_64". In order
to achieve a more compatible tag, you would need..."
- `auditwheel repair` on the already-repaired wheel says:

> cannot repair "wheelhouse/pygame-2.0.0.dev0-cp37-cp37m-manylinux1_x86_64.whl" to "manylinux1_x86_64" ABI because of the presence of too-recent versioned symbols. You'll need to compile the wheel on an older toolchain.

I understand that it's necessary to specify the tag manually at the command line, but it shouldn't successfully apply the `manylinux1` tag if the wheel is not `manylinux1` compatible.

Educated guesswork: I think it checks the validity of the tag based on the files already in the wheel. These are presumably valid for manylinux1. But then it grafts some libraries into the wheel, and these make it invalid.

jjhelmus:
@takluyver Would it be possible to provide the pygame wheel so that anyone (me included) who wanted to debug this issue would have something to test with?

takluyver:
Of course. Though because it's tricky to run auditwheel the first time without the libraries it's built against, maybe it's easier if I push my work into a branch and point you to that?

I'm also investigating a partial fix, but more eyes on it are definitely welcome. :slightly_smiling_face:

takluyver:
OK, the branch is [here](https://github.com/takluyver/pygame/tree/manylinux2010). The last commit works around this issue, but if you step back to https://github.com/takluyver/pygame/commit/61c381b8f72cafc1f9ba555cca9ad149d62997cc you should be able to reproduce it.

From the repo root:

```
cd buildconfig/manylinux-build
make base-image-x64
make wheels-x64
```

takluyver:
#148 is a partial fix - it makes auditwheel error after producing the wheel. I've tested this on the pygame wheels. That's all the time I've got to dig into this today, so if you want to work on a proper fix, go for it.

PR #148, opened 2019-04-11
Check compatibility after repairing wheel

takluyver:
Partially addresses #146.

This checks the output wheel from `auditwheel repair` for compatibility with the specified tag. This is not an ideal fix, because the output wheel is produced even if it's invalid. But producing an error means that it will fail in testing & CI environments, which probably mitigates the problem in many situations.

A better fix would involve either checking the libraries as they are grafted in, or producing the wheel initially into a temporary folder, and only copying it to the output directory if it passes the check.

This builds on the branch in #147.

### DIFF
--- a/auditwheel/main_repair.py
+++ b/auditwheel/main_repair.py
@@ -96,7 +96,7 @@ def execute(args, p):
             msg = ('Produced a wheel %r, but it is no longer compatible '
                    'with the %s tag, possibly because grafted libraries '
                    'require more recent symbols.' %
-                   (out_wheel, reqd_tag))
+                   (out_wheel, args.PLAT))
             sys.exit(msg)
 
         logger.info('\nFixed-up wheel written to %s', out_wheel)
--- a/auditwheel/main_repair.py
+++ b/auditwheel/main_repair.py
@@ -2,6 +2,7 @@ from os.path import isfile, exists, abspath, basename
 from .policy import (load_policies, get_policy_name, get_priority_by_name,
                      POLICY_PRIORITY_HIGHEST)
 import logging
+import sys
 
 logger = logging.getLogger(__name__)
 
@@ -90,5 +91,12 @@ def execute(args, p):
                                      lib_sdir=args.LIB_SDIR,
                                      out_dir=args.WHEEL_DIR,
                                      update_tags=args.UPDATE_TAGS)
+
+        elif reqd_tag < get_priority_by_name(analyzed_tag):
+            msg = ('Produced a wheel %r, but it is no longer compatible '
+                   'with the %s tag, possibly because grafted libraries '
+                   'require more recent symbols.' %
+                   (out_wheel, reqd_tag))
+            sys.exit(msg)
 
         logger.info('\nFixed-up wheel written to %s', out_wheel)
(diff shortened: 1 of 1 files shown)


takluyver:
The test failure on 3.5 might be related to something I just fixed in my other PR:
https://github.com/pypa/auditwheel/pull/147/commits/46619cd0df32f9548cbdb6e1b8d6f6f9de6d07c5

I think the test with numpy is currently sensitive to order - it will succeed if it runs with manylinux1 first, and fail if it runs with manylinux2010 first. Dictionary ordering is deterministic from Python 3.6, but random in 3.5.

ehashman:
Please rebase this PR.

takluyver:
Rebased; thanks for merging my other PR. :slightly_smiling_face:

ehashman:
Looks good @takluyver; is it possible to add a test case to verify this?

takluyver:
I've just been trying to write a test for it, but unfortunately I can't find a good way to do it.

I ran into this with wheels of pygame, but pygame has a number of extra dependencies which need to be built before it, which don't seem reasonable to deal with as part of the test process here. Pygame actually has its own build images adding these dependencies to the standard PyPA images, but there's no published manylinux2010 version yet.

The other packages I've tried to build to test this either fail the check before building (i.e. their compiled libraries already need newer symbols than manylinux1) or they don't need newer symbols at all, so they really are eligible for a manylinux1 tag. It seems I hit a fairly specific corner case.

I've tried so far with numpy, pygame, tornado and pyzmq.

ehashman:
We could make a minimal test case that does exactly what the error message says: have a simple wheel depend on a prebuilt shared object that has more recent symbols than the platform tag? I don't know if it's worth the marginal effort but I'm always worried about preventing regressions :)

takluyver:
I certainly understand the desire for decent tests to avoid regressions. But I don't really know this domain well enough to put together a minimal test case: I don't know how I'd make an extension module only using the older symbols, a separate library using newer symbols, and get them appropriately linked so auditwheel bundles the library in.

ehashman:
If you are interested in learning how to do this, this test is a good example: https://github.com/pypa/auditwheel/blob/b0fde3d6a365aa3f942c9e04ef5bca9014c788ed/tests/test_manylinux.py#L192-L193

The source code for the test package lives here: https://github.com/pypa/auditwheel/tree/b0fde3d6a365aa3f942c9e04ef5bca9014c788ed/tests/testpackage

To get a library with newer symbols, you could just commit a known policy-breaking .so in the git tree and depend on it (rather than building the C code in the manylinux image).


Anyways, if that's too much work, I can look into merging this as is.

takluyver:
Thanks! I'll try to look into this, but I can't guarantee when I'll have time.

mayeut:
@takluyver,
I worked on a test case for the best guess behavior in #146
With dependencies on `manylinux2010` symbols in the "parent"/"main" module, it's rejected when trying to repair it targeting `manylinux1`.
With dependencies on `manylinux2010` symbols only in the "child" module, it's not rejected when trying to repair it targeting `manylinux1`.
You can cherry-pick the commit in my branch in order to add a test in this PR if you find it to fit your needs: https://github.com/mayeut/auditwheel/commit/b3d1ea6c403925e80bc48f0413bc66dbc14fd143
You can see the corresponding failed test here: https://travis-ci.com/mayeut/auditwheel/builds/108599087

PR #150, opened 2019-04-20
 Check external dependencies versioned symbols

### LINKED ISSUE #146: Repair can produce an invalid manylinux1 wheel

mayeut:
This should fix #146 
Uses a different mechanism compared to #148, it checks before producing the wheel.
It checks external dependencies versioned symbols when first analyzing the wheel ABI.


auditwheel/wheel_abi.py
```diff
@@ -112,7 +113,38 @@ def analyze_wheel_abi(wheel_fn: str):
     log.debug('external reference info')
     log.debug(json.dumps(external_refs, indent=4))
 
+    # get versioned symbols from external_refs
+    # since linux tag (priority 0) has no white-list, do not analyze it
+    external_ref_no_linux = dict(set([(v, k) for e in external_refs.values() if e['priority'] != 0 for k, v in e['libs'].items() if v is not None]))
+    external_refs_symbol_policy = {}
+    log.debug(external_ref_no_linux)
+    for path, elf in elf_file_filter([path for path in iter(external_ref_no_linux)]):
```
takluyver: `[path for path in iter(...)]` can be spelled `list(...)`. Or `.keys()` would work, I think.

tests/test_manylinux.py
```diff
@@ -225,6 +225,46 @@ def test_build_wheel_with_binary_executable(docker_container):
...
+
+    # Repair the wheel using the appropriate manylinux container
+    repair_command = (
+        'auditwheel -v repair --plat manylinux1_x86_64 -w /io /io/{orig_wheel}'
+    ).format(policy=policy, orig_wheel=orig_wheel)
+    try:
+        docker_exec(manylinux_id, ['bash', '-c', 'LD_LIBRARY_PATH=/auditwheel_src/tests/testdependencies ' + repair_command])
+        assert policy == 'manylinux1'
+    except CalledProcessError:
+        assert policy != 'manylinux1'
```
takluyver: I follow how the test works here, but the logic is kind of inside out. We're really testing that it succeeds when built on the manylinux1 image but not when built on the manylinux2010 image - because then it uses a newer symbol and isn't manylinux1 compatible.

But what we actually assert is that if repairing works, then it must be a manylinux1 image in use, and if it fails, it must *not* be. The dependent and independent variables are swapped around.

That's OK, and I can't immediately see a neat way to write the test without doing that. But let's have an explanatory comment, and maybe a helpful message if the assertions fail.

mayeut: > But let's have an explanatory comment

I realize now that the comments does not reflect what's done...
I will have a hard time explaining what's done with decent english but let's try. If you have a better suggestion, please do tell. If you think that's good enough, I'll update the commit.
```
# Always try to repair the wheel targeting manylinux1 policy

# with_dependency == 0
# The python module has no dependencies that should be grafted-in and
# uses versioned symbols not available on policies pre-dating the policy matching the image being tested.

# with_dependency == 1
# The python module itself does not use versioned symbols but has a dependency that should be grafted-in that
# uses versioned symbols not available on policies pre-dating the policy matching the image being tested.

# Since manylinux1 has no pre-dating policy, it's the only image/policy that should be able to repair those wheels.
```

> The dependent and independent variables are swapped around.

Maybe something like that would work:
```
repair_command = (
	        'LD_LIBRARY_PATH=/auditwheel_src/tests/testdependencies auditwheel -v repair --plat manylinux1_x86_64 -w /io /io/{orig_wheel}'
	    ).format(policy=policy, orig_wheel=orig_wheel)
if policy != 'manylinux1':
    with self.assertRaises(CalledProcessError, msg='Repairing this wheel targeting manylinux1  shall fail on {policy} image'.format(policy=policy)):
        docker_exec(manylinux_id, ['bash', '-c', repair_command])
    return

# go-on checking all works properly on manylinux1 image
docker_exec(manylinux_id, ['bash', '-c', repair_command])
...
```

tests/test_manylinux.py
```diff
@@ -225,6 +225,46 @@ def test_build_wheel_with_binary_executable(docker_container):
...
+    assert len(filenames) == 2
+    repaired_wheels = [fn for fn in filenames if policy in fn]
+    # Wheel picks up newer symbols when built in manylinux2010
+    expected_wheel_name = 'testdependencies-0.0.1-cp35-cp35m-%s_x86_64.whl' % policy
+    assert repaired_wheels == [expected_wheel_name]
+    repaired_wheel = repaired_wheels[0]
+    output = docker_exec(manylinux_id, 'auditwheel show /io/' + repaired_wheel)
+    assert (
+        'testdependencies-0.0.1-cp35-cp35m-{policy}_x86_64.whl is consistent'
+        ' with the following platform tag: "{policy}_x86_64"'
```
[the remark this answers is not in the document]
mayeut: yes, good point.

tests/test_manylinux.py
takluyver: I think the rearrangement you propose is an improvement - it's more verbose, but it makes the intent more clear.

I'm happy to have a go at clarifying any comments I can understand, as I am a native English speaker. :slightly_smiling_face:

mayeut: Should be better now. I force-pushed a new commit for tests.

tests/test_manylinux.py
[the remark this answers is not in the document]
mayeut: Now done. I force-pushed a new commit for tests.

auditwheel/wheel_abi.py
```diff
@@ -112,7 +113,38 @@ def analyze_wheel_abi(wheel_fn: str):
     log.debug('external reference info')
     log.debug(json.dumps(external_refs, indent=4))
 
+    # get versioned symbols from external_refs
+    # since linux tag (priority 0) has no white-list, do not analyze it
+    external_ref_no_linux = dict(set([(v, k) for e in external_refs.values() if e['priority'] != 0 for k, v in e['libs'].items() if v is not None]))
```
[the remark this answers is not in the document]
mayeut: I pushed a new commit that I hope clarify things a bit. I'll squash it in the previous commit if that works for you otherwise, let's iterate a bit more before doing that.

auditwheel/wheel_abi.py
```diff
@@ -112,7 +113,46 @@ def analyze_wheel_abi(wheel_fn: str):
...
+    for e in external_refs.values():  # go through all policies
+        if e['priority'] == 0:
+            continue    # since linux tag (priority 0) has no white-list, do not analyze it
+        for libname, realpath in e['libs'].items():  # go through all referenced libs, retrieving their name and realpath
+            if realpath is None:
+                continue  # we can't analyze the lib if we don't know where it's located
+            if (realpath, libname) not in external_ref_no_linux:
+                external_ref_no_linux.append((realpath, libname))
+    external_ref_no_linux = dict(external_ref_no_linux)  # create a dict using the realpath as a key, libname as a value
+    external_refs_symbol_policy = {}  # create a dict using libname as a key, versioned symbols as a value
```
takluyver: A comment like this could more concisely be written: `# {libname: versioned_symbols}`

But it's still a bit hard to follow what's going on. Can we expand a bit more what versioned symbols are?

```python
external_refs_symbol_policy = {}  # {libname: {???: set([???]) }}
```

It might also be useful to give some example values alongside the prototype:

```python
# {'libc.so.6': {???: set([???]) }}
```

tests/test_manylinux.py
```diff
@@ -225,6 +225,66 @@ def test_build_wheel_with_binary_executable(docker_container):
...
+
+    repair_command = (
+        'LD_LIBRARY_PATH=/auditwheel_src/tests/testdependencies auditwheel -v repair --plat manylinux1_x86_64 -w /io /io/{orig_wheel}'
+    ).format(policy=policy, orig_wheel=orig_wheel)
+
+    if policy != 'manylinux1':
+        # we shall fail on non manylinux1 images
+        with pytest.raises(CalledProcessError):
+            docker_exec(manylinux_id, ['bash', '-c', repair_command])
+    return
```
takluyver: I think the indentation is wrong here, so the rest of the test will never run.

tests/test_manylinux.py
```diff
@@ -225,6 +225,66 @@ def test_build_wheel_with_binary_executable(docker_container):
...
+def test_build_wheel_with_highest_policy_and_image_dependencies(with_dependency, docker_container):
+    # Always try to repair the wheel targeting manylinux1 policy
+    #
+    # with_dependency == 0
+    #   The python module has no dependencies that should be grafted-in and
+    #   uses versioned symbols not available on policies pre-dating the policy matching the image being tested.
+    # with_dependency == 1
+    #   The python module itself does not use versioned symbols but has a dependency that should be grafted-in that
+    #   uses versioned symbols not available on policies pre-dating the policy matching the image being tested.
+    # Since manylinux1 has no pre-dating policy, it's the only image/policy that should be able to repair those wheels.
```
takluyver: This sentence doesn't quite match my understanding of how the test works.

We can repair the wheel built on any image, but only to the target platform corresponding to the image. The test checks that a wheel built on manylinux2010 can be repaired to target manylinux1, and that a wheel built on manylinux1 can be.

auditwheel/wheel_abi.py
```diff
@@ -112,7 +113,46 @@ def analyze_wheel_abi(wheel_fn: str):
...
+    external_ref_no_linux = []
+    for e in external_refs.values():  # go through all policies
+        if e['priority'] == 0:
+            continue    # since linux tag (priority 0) has no white-list, do not analyze it
+        for libname, realpath in e['libs'].items():  # go through all referenced libs, retrieving their name and realpath
+            if realpath is None:
+                continue  # we can't analyze the lib if we don't know where it's located
+            if (realpath, libname) not in external_ref_no_linux:
+                external_ref_no_linux.append((realpath, libname))
+    external_ref_no_linux = dict(external_ref_no_linux)  # create a dict using the realpath as a key, libname as a value
```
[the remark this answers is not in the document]
mayeut: nope, I just transformed the one-liner without thinking... will change that.

tests/test_manylinux.py
mayeut: right, will correct

mayeut:
@takluyver, I pushed 2 commits that should answer your comments

takluyver:
Thanks, the comments really do help me understand what's going on.

There are some rather long lines there now. I'm not picky about any particular column limit, but some of these lines are definitely longer than is comfortably readable. Do you have time to reformat the code a bit? Or I can have a go at this if you prefer.

takluyver:
Now #158 added a linter to the test suite, so we will need to stick to some column limit to satisfy that.

mayeut:
I’ll rebase next week in order to get the linter checks

mayeut:
I've rebased this PR on master, it now passes the lint check but I'm not quite happy with the type description of some variables. I don't know how to improve the readability of those comments, so feel free to comment on these.

auditwheel/wheel_abi.py
```diff
@@ -116,7 +117,65 @@ def analyze_wheel_abi(wheel_fn: str):
...
+        # go through all libs, retrieving their soname and realpath
+        for libname, realpath in e['libs'].items():
+            # we can't analyze the lib if we don't know where it's located
+            if realpath is None:
+                continue
+            if realpath not in external_ref_no_linux.keys():
+                external_ref_no_linux[realpath] = libname
+    # {libname: {depname: set([symbol_version])}} e.g.
+    # {'external_ref.so.1': {'libc.so.6', set(['GLIBC_2.5','GLIBC_2.12'])}}
+    external_refs_symbol_policy = {}
```
lkollar: This should be another function, returning the `external_refs_symbol_policy` dict.

auditwheel/wheel_abi.py
```diff
@@ -116,7 +117,65 @@ def analyze_wheel_abi(wheel_fn: str):
...
+    for path, elf in elf_file_filter(external_ref_no_linux.keys()):
+        # {depname: set(symbol_version)}, e.g.
+        # {'libc.so.6', set(['GLIBC_2.5','GLIBC_2.12'])}
+        elf_versioned_symbols = defaultdict(lambda: set())
```
lkollar: `defaultdict(set)` should work.

auditwheel/wheel_abi.py
```diff
@@ -116,7 +117,65 @@ def analyze_wheel_abi(wheel_fn: str):
...
+        for key, value in elf_find_versioned_symbols(elf):
+            log.debug('path %s, key %s, value %s', path, key, value)
+            elf_versioned_symbols[key].add(value)
+        external_refs_symbol_policy[external_ref_no_linux[path]] = \
```
lkollar: This would be more readable as two separate assignments and it would probably fit without a break.

```
libname = external_ref_no_linux[path]
external_refs_symbol_policy[libname] = elf_versioned_symbols
```

auditwheel/wheel_abi.py
```diff
@@ -116,7 +117,65 @@ def analyze_wheel_abi(wheel_fn: str):
...
+            elf_versioned_symbols
+
     symbol_policy = versioned_symbols_policy(versioned_symbols)
+
+    # since white-list is different per policy,
+    # let's examine versioned_symbol per policy when including external refs
+    # list of tuples of the form (policy_priority, versioned_symbols), e.g.
+    # [(100, {'libc.so.6', set(['GLIBC_2.5'])})]
+    symbol_policies = []
```
lkollar: This should be a separate function too, returning `symbol_policies`.

auditwheel/wheel_abi.py
```diff
@@ -116,7 +117,65 @@ def analyze_wheel_abi(wheel_fn: str):
     log.debug('external reference info')
     log.debug(json.dumps(external_refs, indent=4))
 
+    # get versioned symbols from external_refs
+    # {realpath: libname} e.g.
+    # {'/path/to/external_ref.so.1.2.3': 'external_ref.so.1'}
+    external_ref_no_linux = {}
+    # go through all policies
+    for e in external_refs.values():
```
lkollar: I would much prefer proper variable names in the loops instead of single letter names. I'm surprised the linter doesn't complain about this.

tests/test_manylinux.py
```diff
@@ -225,6 +226,63 @@ def test_build_wheel_with_binary_executable(docker_container):
         python_id, ['python', '-c', 'from testpackage import runit; print(runit(1.5))']).strip()
     assert output.strip() == '2.25'
 
+@pytest.mark.parametrize('with_dependency', ['0', '1'])
+def test_build_wheel_with_highest_policy_and_image_dependencies(with_dependency, docker_container):
```
lkollar: Can you try to make the line lengths shorter to make it more consistent with the rest of the file? We don't run the linter on tests at the moment but the rest of the tests are mostly under 80 characters.

lkollar: I really like this test and the native extension.

tests/test_manylinux.py
```diff
@@ -225,6 +226,63 @@ def test_build_wheel_with_binary_executable(docker_container):
...
+    # with_dependency == 0
+    #   The python module has no dependencies that should be grafted-in and
+    #   uses versioned symbols not available on policies pre-dating the policy matching the image being tested.
+    # with_dependency == 1
+    #   The python module itself does not use versioned symbols but has a dependency that should be grafted-in that
+    #   uses versioned symbols not available on policies pre-dating the policy matching the image being tested.
+
+    policy, manylinux_id, python_id, io_folder = docker_container
+
+    docker_exec(manylinux_id, ['bash', '-c', 'cd /auditwheel_src/tests/testdependencies && WITH_DEPENDENCY={} python setup.py -v build_ext -f bdist_wheel -d /io'.format(with_dependency)])
```
lkollar: Wouldn't `pip wheel` work here? It would be simpler. There is an example in `test_build_wheel_with_binary_executable` for this.

mayeut: I can't use `pip wheel` because I need to force rebuilding the extension between tests (otherwise the extension used in a previous test will be reused without the correct `WITH_DEPENDENCY` option)

tests/test_manylinux.py
mayeut: done

auditwheel/wheel_abi.py
```diff
@@ -116,7 +117,65 @@ def analyze_wheel_abi(wheel_fn: str):
...
+        # linux tag (priority 0) has no white-list, do not analyze it
+        if e['priority'] == 0:
+            continue
+        # go through all libs, retrieving their soname and realpath
```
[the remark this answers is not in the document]
takluyver: Sorry, I was the one pushing for more comments because I didn't understand the code, and it's possible that I pushed a bit too much.

lkollar:
@auvipy I was going to review the changes. Why did you merge this? In the future could you wait until the person who requested changes approves the final version?

auvipy:
OK sure.

PR #151, opened 2019-04-22
Use AUDITWHEEL_PLAT environment variable as a default option to --plat

mayeut:


mayeut:
I'll add that

mayeut:
I've reworked that a bit with inspiration from how twine does it.
I'll add more tests once #150 is merged-in (or at least the test commit)

tests/helpers.py
```diff
@@ -0,0 +1,37 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Test functions useful across twine's tests."""
+
+import contextlib
+import os
+
+
+@contextlib.contextmanager
+def set_env(**environ):
```
takluyver: pytest's [monkeypatch](https://docs.pytest.org/en/latest/monkeypatch.html) fixture can be used to modify an environment variable for a test - up to you, but it saves an extra helper function.

mayeut: Done

takluyver:
@mayeut I'm happy to merge this. You mentioned before you were waiting for #150. Do you want to get that one in first, in case there are any odd interactions between them? It seems unlikely, but if there are, I guess it's easy to tidy them up in this simpler change.

mayeut:
I'll add some more tests (integration) in a different PR

Issue #154, opened 2019-05-01
Remove non-Linux platform restriction


lkollar:
Right now auditwheel exits on startup on non-Linux systems. There is probably little use in running it on other systems, but this also makes it difficult to develop on these platforms as it causes all test cases to fail. Given that all platform-specific tests run in Docker, development on non-Linux platforms should be perfectly doable.

I think this would be useful to make it easier to contribute to the project.

After removing the check I can run the test suite successfully on macOS.

https://github.com/pypa/auditwheel/blob/b0fde3d6a365aa3f942c9e04ef5bca9014c788ed/auditwheel/policy/__init__.py#L29-L34

PR #155, opened 2019-05-01
Make it possible to run the test suite on non-Linux systems

lkollar:
At the moment the platform check is done when the policy module gets
imported. This makes it difficult to run the test suite on non-Linux
platforms. Moving the check to the CLI entry point allows the tests to
run and retains the current behavior and exits on non-Linux platforms
when running the tool.

Note that in a previous commit I accidentally commented out the check,
that's why it's shown commented in this diff.

Closes: #154

### DIFF
--- a/auditwheel/main.py
+++ b/auditwheel/main.py
@@ -11,6 +11,10 @@ from . import main_repair
 
 
 def main():
+    if sys.platform != 'linux':
+        print('Error: This tool only supports Linux')
+        return 1
+
     dist = pkg_resources.get_distribution('auditwheel')
     version = 'auditwheel %s installed at %s (python %s)' % (
         dist.version, dist.location, sys.version[:3])
--- a/auditwheel/policy/__init__.py
+++ b/auditwheel/policy/__init__.py
@@ -27,13 +27,6 @@ _PLATFORM_REPLACEMENT_MAP = {
     'manylinux2010_i686': ['linux_i686'],
 }
 
-# XXX: this could be weakened. The show command _could_ run on OS X or
-# Windows probably, but there's not much reason to inspect foreign package
-# that won't run on the platform.
-# if platform != 'linux':
-#     logger.critical('Error: This tool only supports Linux')
-#     sys.exit(1)
-
 
 def get_arch_name():
     if _platform_module.machine() in non_x86_linux_machines:


auditwheel/policy/__init__.py
```diff
@@ -26,12 +26,11 @@
     'manylinux2010_i686': ['linux_i686'],
 }
 
-# XXX: this could be weakened. The show command _could_ run on OS X or
-# Windows probably, but there's not much reason to inspect foreign package
-# that won't run on the platform.
 if platform != 'linux':
-    logger.critical('Error: This tool only supports Linux')
-    sys.exit(1)
+    logger.warning(f'Running on {platform}. Note that this tool only '
```
mayeut: I'm no expert on this but it seems the syntax used for logging here is only supported on python 3.6+ but not python 3.5
Can you update this so that tests do pass please ?

lkollar: Oops, sorry. I forgot that 3.5 is supported. Updated.

lkollar:
@mayeut sorry, but I just realised that this should use `printp` instead of the logger, just like all other text displayed to users. This will also wrap the text correctly. Can you please review it again?

auvipy:
@mayeut I would like to take some more time to check this in depth

lkollar:
@auvipy: I rebased it on top of master. To your question: on macOS the original restriction made all tests fail, but with this change the entire test suite passes. I expect this would be the same on any platform where the `docker` executable is available and the test dependencies are present.

auditwheel/policy/__init__.py
```diff
@@ -27,12 +28,10 @@
...
 }
 
-# XXX: this could be weakened. The show command _could_ run on OS X or
-# Windows probably, but there's not much reason to inspect foreign package
-# that won't run on the platform.
-# if platform != 'linux':
-#     logger.critical('Error: This tool only supports Linux')
-#     sys.exit(1)
+if platform != 'linux':
+    printp('Running on {}. Note that this tool only '
```
auvipy: I guess wording needs some adjustment than...
" Note that this tool was designed to work mainly on Linux systems, So results on Non-Linux platforms might be incorrect"

lkollar:
We'd need to set up non-Linux build bots for this, but given that this still only means that we don't just bomb out on those platforms there isn't much to validate. Running `auditwheel` on non-Linux systems is still not very useful as many things will just not work (thus the warning in this PR). I'd say running on non-Linux is completely unsupported.

auditwheel/policy/__init__.py
lkollar: Do you mean that we should say that `auditwheel` can be used on non-Linux systems? Apologies if I misunderstood your comment but it would be helpful if you could elaborate on why you think this change in wording is necessary.

auvipy: No NO, just to make the sentence more softer... :D

auvipy: you can adjust as your wish

auvipy: If you want I can get this in as is

takluyver:
How far does it get if you actually run `auditwheel show` or `auditwheel repair` on a wheel file on a non-Linux platform? Does it crash at some later stage?

I'm fine with making it possible to run the tests on non-Linux platforms, but if it can actually run and give incorrect results on a Mac or BSD system, then I think it's worth having a bit more than an easily ignored warning.

Maybe we can move the check from import time to the main function for the command-line interface, so it still fails if you try to run `auditwheel` at a shell.

auditwheel/policy/__init__.py
auvipy: I would love to hear your responses on the queries of Thomas's

lkollar:
@takluyver `auditwheel` runs fine on macOS, it just fails to report some information, like the symbol versions it finds in the wheel. `repair` doesn't work at all. I do agree that it's safer to keep the check as it is, as there might be situations where the reported information can be confusing otherwise.

Updated the PR to move the check the CLI entry point.

auditwheel/main.py
```diff
@@ -11,6 +11,10 @@
 
 
 def main():
+    if sys.platform != 'linux':
```
takluyver: It's recommended to use `.startswith('linux')` for this check - docs: https://docs.python.org/3/library/sys.html#sys.platform

lkollar: That should only be necessary for versions older than 3.3, which we don't support (see the last paragraph in the section you linked to). I'm happy to make the change if you feel strongly about it though.

auvipy: > That should only be necessary for versions older than 3.3, which we don't support (see the last paragraph in the section you linked to). I'm happy to make the change if you feel strongly about it though.

It means from python 3.3 and on startswith linux only preferred, not it's only supported in python 3.3!!

lkollar: I meant that `auditwheel` only runs on Python 3.5 and above.

takluyver: Fair enough. I had forgotten that auditwheel doesn't aim for Python 2 support.

takluyver:
Feel free to merge once CI passes.

PR #156, opened 2019-05-02
Improve error message on invalid package

lkollar:
Instead of raising an empty ValueError when encountering an unknown file format, provide a meaningful error message in the exception. This should make it easier for users to tell what went wrong.


takluyver:
Maybe the error message can go even more descriptive, and point out that it recognises .whl and .tar.bz2 extensions?

Also, interesting, I hadn't noticed that auditwheel had code for working with conda packages!

PR #157, opened 2019-05-03
Use pyproject.toml

lkollar:
Make setuptools a build dependency in pyproject.toml and set it as the
build backend.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,3 @@
+[build-system]
+requires = ["setuptools", "wheel", "pbr"]
+build-backend = "setuptools.build_meta"
--- a/tests/test_manylinux.py
+++ b/tests/test_manylinux.py
@@ -195,7 +195,7 @@ def test_build_wheel_with_binary_executable(docker_container):
     policy, manylinux_id, python_id, io_folder = docker_container
     docker_exec(manylinux_id, 'yum install -y gsl-devel')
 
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[build-system]
+requires = ["setuptools", "pbr"]
```
pganssle: Does this need `wheel` as well?

tests/test_manylinux.py
```diff
@@ -195,7 +195,7 @@ def test_build_wheel_with_binary_executable(docker_container):
     policy, manylinux_id, python_id, io_folder = docker_container
     docker_exec(manylinux_id, 'yum install -y gsl-devel')
 
-    docker_exec(manylinux_id, ['bash', '-c', 'cd /auditwheel_src/tests/testpackage && python setup.py bdist_wheel -d /io'])
+    docker_exec(manylinux_id, ['bash', '-c', 'cd /auditwheel_src/tests/testpackage && python -m pip wheel -w /io .'])
```
pganssle: If you are going to use `pip` to build wheels, it's best to also use `--no-deps`.

I personally prefer `pep517.build`.

lkollar: As this is only invoked for for a test inside a container, I'd leave `pip` here to keep it simple. Adding `--no-deps`.

lkollar:
@auvipy given that this is a relatively small change and has already been approved by Paul, I'm going to land it.

PR #165, opened 2019-05-24
Fix typos in docstrings and info messages

jcfr:


jcfr:
Thanks for reviewing and integrating so quickly :+1:

Issue #163, opened 2019-05-17
Add unit tests to the project


lkollar:
At the moment a large portion if the code base is not covered with unit tests. There are quite a few integration or functional tests which run the entire application to validate the overall behaviour, but ideally we should have full unit test coverage around the whole codebase.

Once this is done we can enable a coverage service on the repo to make sure the coverage level doesn't drop during contributions.

I plan to keep this issue to track the PRs needed to add unit tests. It's probably best to break this into separate contributions to ease reviews.

Files mentioned in #126 should be covered by the vendored projects' own unit tests.

lkollar:
Let's reopen this as I intend to keep for the rest of the unit tests.

lkollar:
I wanted to address @takluyver's [request](https://github.com/pypa/auditwheel/pull/169#issuecomment-503127058) on the overall plan for testing in this repo as part of this issue.

I should have probably better explained what my intention was with the plan to add unit test coverage. When I first started contributing to `auditwheel` I found it difficult to validate the changes I made as the test suite is very high-level, for the most part exercising the entire application -- it is quite difficult to make a small change to a part of the application and test just that part in isolation, rather than spinning up Docker containers and installing stuff with `pip`. It should be much easier to validate simple changes and I think that adding unit tests will help contributions as a result.

I'm sure the other maintainers are also aware of concept of the [testing pyramid](https://martinfowler.com/articles/practical-test-pyramid.html) but just for completeness I wanted to sum up the idea here: separating higher level tests which exercise the entire application from unit tests which validate individual components provide multiple benefits. The most important one in my view is that it speeds up the feedback cycle when working on the code. Given that a large part of `auditwheel`'s test suite runs in Docker and already takes a long time to run I think it is very important to start creating unit tests which run faster and make debugging test failures easier.

[...]

PR #166, opened 2019-06-04
test: Add unit test for hashfile

lkollar:
More tests for #163.

### DIFF
--- a/tests/unit/test_hashfile.py
+++ b/tests/unit/test_hashfile.py
@@ -1,0 +1,14 @@
+from io import BytesIO
+
+from auditwheel.hashfile import hashfile
+
+
+def test_hash():
+    # GIVEN
+    mock_file = BytesIO(b"this is a test file")
+
+    # WHEN
+    result = hashfile(mock_file)
+
+    # THEN
+    assert result == "5881707e54b0112f901bc83a1ffbacac8fab74ea46a6f706a3efc5f7d4c1c625"


takluyver:
This feels vastly overcomplicated for testing a straightforward function.

- The hash function is mocked out, so we're not actually checking that it produces the correct SHA256 hash.
- The input is invalid - it expects a binary file object, but the mocking obscures that.
- `blocksize=0` would produce an incorrect result in real life, so it doesn't make sense to validate the current behaviour with a test.

Hash algorithms have to have entirely predictable output. So there's no need to mock it out. If we really need a test for this, just produce some data (`b"blah " * 20000`) and check that the result matches a pre-calculated hash.

lkollar:
My intention is to build up 100% (or close to that) unit test coverage around the current code base. Ideally unit tests should test the system under test only without invoking any external functionality. I agree that the current test is a bit too complicated though.

> * The hash function is mocked out, so we're not actually checking that it produces the correct SHA256 hash.

The intent here is not to validate the SHA256 algorithm but to test the `hashfile` function itself. Testing that SHA256 hashing works is the responsibility of the test suite around `hashfile`. That said, I agree that it would be more practical to only mock the input file and still call into `hashlib`. That way the test becomes simpler and given that `hashlib` is in the standard library and does not involve using disk or network it should be fine to call it.

> * `blocksize=0` would produce an incorrect result in real life, so it doesn't make sense to validate the current behaviour with a test.

The function does not disallow `blocksize=0` at the moment so I wanted to add a test to make sure the behaviour is documented. I can remove it though, I don't think this is important.

I will simplify this by passing a bytes-like object and remove the mock for `hashlib`.

takluyver:
Thanks, that looks sensible now.

I'll push back on over-testing things like `blocksize=0` because it adds friction to working on the codebase if you have to go and understand and fix tests which are failing for something no-one cares about. Testing and coverage measurements are tools, not goals in themselves.

PR #167, opened 2019-06-15
Break up docker test fixtures

takluyver:
The integration tests previously had a single big `docker_container` fixture which actually started 2 docker containers. This breaks it up into separate fixtures, which simplifies things a bit, and allows tests to use one container without starting the other (one test already does so).

It also switches from constructing `docker` commands to run in a subprocess to using the docker Python API.

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -5,3 +5,4 @@ numpy
 pypatchelf
 flake8
 pretend
+docker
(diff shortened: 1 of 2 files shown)


takluyver:
Test duration appears to be similar at worst, and maybe shaves off ~30 seconds.

takluyver:
Thanks. I'll give it another day or two for other people to have a look, then merge it if there are no other comments.

Issue #142, opened 2019-03-21
Wheelhouse is 404


di:
The `WHEELHOUSE=1` Travis test currently tries to fetch wheelhouse from https://github.com/getnikola/wheelhouse/archive/v2.7.zip, however this link is 404 and the repo is missing as well: https://github.com/getnikola/wheelhouse

This test is currently an allowed failure, do we need it? Do we need to update the URL for wheelhouse?

ehashman:
The PRs you'd want for context are here: #86 #87

If someone wants to update this test to make it stop failing, I'd be happy to review. We'd need to swap out the nonexistent wheelhouses with ones that are still live.

lkollar:
The tests have been removed in #179.

Issue #159, opened 2019-05-04
Python C++ module not properly repaired, retains dependency on symlink to bundled library


madig:
I have a Python C++ module that links to libcrypto from OpenSSL and libcurl. I build the package on the manylinux2010 container (`python3 setup.py bdist_wheel`) and run `auditwheel repair` on the .whl. The library shared objects (libcrypto, libz and libcurl) are correctly packed in and ldd-ing the Python module shared object correctly finds and shows them as dependencies, yet it retains a dependency on "libcrypto.so.10", which is named differently on e.g. Debian-based containers, making import fail. What am I missing?

madig:
Ugh. I can't even run the resulting wheel on the manylinux2010 container, same eror:

```
ImportError: /tmp/venv/lib/python3.7/site-packages/dsign/.libs/libcrypto-a94305d7.so.1.0.1e: version `libcrypto.so.10' not found (required by /tmp/venv/lib/python3.7/site-packages/dsign/_dsign.cpython-37m-x86_64-linux-gnu.so)
```

madig:
I tried linking the library to libcrypto statically and it works, but:
```
(venv) [root@a381fbecf6ea dsign]# auditwheel show dist/dsign-1.4.5.dev0+gf7b8133.d20190506-cp37-cp37m-linux_x86_64.whl

dsign-1.4.5.dev0+gf7b8133.d20190506-cp37-cp37m-linux_x86_64.whl is
consistent with the following platform tag: "linux_x86_64".

The wheel references external versioned symbols in these system-
provided shared libraries: libgcc_s.so.1 with versions {'GCC_3.0'},
libpthread.so.0 with versions {'GLIBC_2.2.5'}, libc.so.6 with versions
{'GLIBC_2.3.4', 'GLIBC_2.7', 'GLIBC_2.4', 'GLIBC_2.3', 'GLIBC_2.2.5'},
libstdc++.so.6 with versions {'CXXABI_1.3', 'GLIBCXX_3.4'}

This constrains the platform tag to "manylinux2010_x86_64". In order
to achieve a more compatible tag, you would need to recompile a new
wheel from source on a system with earlier versions of these
libraries, such as a recent manylinux image.
```
```
(venv) [root@a381fbecf6ea dsign]# auditwheel repair dist/dsign-1.4.5.dev0+gf7b8133.d20190506-cp37-cp37m-linux_x86_64.whl
[...]

PR #168, opened 2019-06-15
Fix running 'wheelhouse' test script on Travis CI

takluyver:
This is an allowed failure, but so long as it's there, it might as well work.

### DIFF
--- a/tests/integration/test-wheelhouses.sh
+++ b/tests/integration/test-wheelhouses.sh
@@ -1,19 +1,8 @@
 set -e
 
-# Download some wheels from one location
-PY_VERS="2.7 3.3 3.4 3.5"
-for PYVER in ${PY_VERS}; do
-    wget "https://github.com/getnikola/wheelhouse/archive/v${PYVER}.zip"
-    unzip "v${PYVER}.zip"
-done
-
 # Download some other wheels that Nathaniel made
-#URL=https://vorpus.org/~njs/tmp/manylinux-test-wheels/original/
-#curl --silent $URL | grep 'a href' | grep whl | cut -d ' ' -f 8 | cut -d '=' -f 2 | cut -d '"' -f 2 | xargs -n1 -I '{}' wget --no-check-certificate -P wheelhouse-njs "$URL/{}"
-
-# Download some more wheels that Robert made
-URL=http://stanford.edu/~rmcgibbo/wheelhouse/
-curl --silent $URL | grep 'a href' | grep whl | cut -d ' ' -f 8 | cut -d '=' -f 2 | cut -d '"' -f 2 | xargs -n1 -I '{}' wget -P wheelhouse-rmcgibbo "$URL/{}" -nc
+URL=https://vorpus.org/~njs/tmp/manylinux-test-wheels/original/
+curl --silent $URL | grep 'a href' | grep whl | cut -d ' ' -f 8 | cut -d '=' -f 2 | cut -d '"' -f 2 | xargs -n1 -I '{}' wget --no-check-certificate -P wheelhouse-njs "$URL/{}"
 
 # These are boring
 rm -f wheelhouse*/*-none-any.whl
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,6 @@ matrix:
     env: WHEELHOUSE=1
   - python: "3.6"
   - python: "3.7"
-  allow_failures:
-  - python: "3.5"
-    env: WHEELHOUSE=1
 
 sudo: required
 services:
--- a/tests/travis.sh
+++ b/tests/travis.sh
@@ -3,7 +3,7 @@
 set -exo pipefail
 
 if [[ "$WHEELHOUSE" == "1" ]]; then
-    bash tests/test-wheelhouses.sh
+    bash tests/integration/test-wheelhouses.sh
 else
     pytest -s
     auditwheel lddtree $(python -c 'import sys; print(sys.executable)')


takluyver:
@njsmith: this change would have the CI hitting your vorpus.org server - the code was there before, but commented out. I'd like your approval before merging it.

@rmcgibbo: your 'wheelhouse' at stanford.edu is no longer available. Let me know if you have the same files at some other URL. I don't think this is high priority, so don't spend energy on it, but if it's just a case of tweaking the URL, we can do that.

rmcgibbo:
I don't have those files anymore (Stanford took down my webspace a couple months after I graduated), but they weren't anything special.

rmcgibbo:
@takluyver: from these [these](https://travis-ci.org/pypa/auditwheel/builds/176840549) travis logs from 3 years ago, it looks to me like the tests were covering wheels from `lxml-3.7.3`, `Pillow-4.1.1`, `PyYAML-3.12`.

takluyver:
Thanks! Do you think it's worth re-creating some more tests with real-world wheels, or is it more a nice-to-have alongside the main test suite? I wanted to make sure that whatever's there is running effectively, but I don't know how important it is.

tests/integration/test-wheelhouses.sh
```diff
@@ -1,19 +1,8 @@
...
-done
-
 # Download some other wheels that Nathaniel made
-#URL=https://vorpus.org/~njs/tmp/manylinux-test-wheels/original/
-#curl --silent $URL | grep 'a href' | grep whl | cut -d ' ' -f 8 | cut -d '=' -f 2 | cut -d '"' -f 2 | xargs -n1 -I '{}' wget --no-check-certificate -P wheelhouse-njs "$URL/{}"
-
-# Download some more wheels that Robert made
-URL=http://stanford.edu/~rmcgibbo/wheelhouse/
-curl --silent $URL | grep 'a href' | grep whl | cut -d ' ' -f 8 | cut -d '=' -f 2 | cut -d '"' -f 2 | xargs -n1 -I '{}' wget -P wheelhouse-rmcgibbo "$URL/{}" -nc
+URL=https://vorpus.org/~njs/tmp/manylinux-test-wheels/original/
```
lkollar: Why not include these wheels in the source repo instead? We already have a few under the tests directory.

lkollar:
This should fix #142. We should also remove the `allow_failures` bit in `.travis.yml`.

tests/integration/test-wheelhouses.sh
takluyver: I wouldn't want to clutter up the source repo with various multi-MB wheels; large binary files are a bit awkward to handle in git. This test can also easily be expanded to many more wheels, e.g. I saw [this scikit-learn wheelhouse](https://3f23b170c54c2533c070-1c8a9b3114517dc5fe17b7c3f8c63a43.ssl.cf2.rackcdn.com/) linked from #86.

takluyver:
The latest commit removes the `allow_failures` section.

I'm in two minds about this. On the one hand, it could be a flaky test, as it relies on downloading extra files. But then I don't think `allow_failures` tests are much use. At least the way I use Travis, I usually ignore them, and as these have been failing for about 18 months, it seems I'm not the only one.

In any case, I'd still like an OK from @njsmith before we merge this, because it will hit his server.

tests/integration/test-wheelhouses.sh
mayeut: One way not to clutter up the source repo with various multi-MB wheels would be to use `git-lfs`.
> Do you think it's worth re-creating some more tests with real-world wheels, or is it more a nice-to-have alongside the main test suite? I wanted to make sure that whatever's there is running effectively, but I don't know how important it is.

Well, I think that's the main question here. I don't know if it's important either... Were those tests added to cover a feature, bug, in the past, I can't say looking at the history...

mayeut:
> Do you think it's worth re-creating some more tests with real-world wheels, or is it more a nice-to-have alongside the main test suite? I wanted to make sure that whatever's there is running effectively, but I don't know how important it is.

Well, I think that's the main question here. I don't know if it's important either... Were those tests added to cover a feature, bug, in the past, I can't say looking at the history...

rmcgibbo:
I added those tests, but it was 3 years ago so I don't remember my exact train of thought. Generally though, at that time this project was something I was just hacking on. I don't think it was an official pypa project, I don't think that the first manylinux pep had yet been merged (it was probably under discussion), and it wasn't clear whether this software would even fill the intended use case. There wasn't really a "main test suite". I think I basically figured "if I can process some random wheels that I found on the internet, then it's probably working, and that's easier than investing in writing a bunch of unit tests right now"

So... real-world tests are always great. But I'd recommend making the decision on whether or not to include tests like this based on your current assessment of the coverage of the test suite, not some no-longer-applicable-determination I made three years ago when this was a solo hobby project that I was hacking on while procrastinating on my dissertation.

takluyver:
Thanks @rmcgibbo. That's the sort of scenario I would have guessed that script came from. :-)

@lkollar is doing some good work towards increasing test coverage. But I don't yet have a good feel for how well the test suite exercises gnarly cases that might come up in real-world wheels.

lkollar:
Given what @rmcgibbo said it might not be useful to try to fix these tests with the same wheels. The `test-wheelhouse.sh` script does not try to import the repaired wheels either, so this test is not that useful anyway. The repair step's return value is not even checked so we won't catch issues there. We now have several more exhaustive tests which validate specific features and actually test that the resulting libraries are usable. So I think we should consider two alternatives here:

- Remove these tests altogether and keep adding integration tests for wheels for the gnarly cases.  https://github.com/pypa/auditwheel/issues/159  is an example but we would have to xfail it until `patchelf` lands the fix. I'm sure we could find more in the issue tracker.
- Keep the wheels but replace the shell script with a proper Python test which repairs and imports the modules. 

If we keep the wheel I still think that we should store them in the source repo. [pip](https://github.com/pypa/pip/tree/master/tests/data/packages) does this too and it seems to be working just fine.

takluyver:
Now we've got coverage measurement set up and we can see we're exercising most of the code already, I think I'm comfortable with removing the wheelhouses script, and #179 does that.

takluyver:
#179 has been merged instead of this.

PR #169, opened 2019-06-17
Add unit tests for `main.py`

lkollar:
More tests for #163.

### DIFF
--- a/tests/unit/test_main.py
+++ b/tests/unit/test_main.py
@@ -1,0 +1,29 @@
+import sys
+
+from auditwheel.main import main
+
+
+def test_unsupported_platform(monkeypatch):
+    # GIVEN
+    monkeypatch.setattr(sys, "platform", "unsupported_platform")
+
+    # WHEN
+    retval = main()
+
+    # THEN
+    assert retval == 1
+
+
+def test_help(monkeypatch, capsys):
+    # GIVEN
+    monkeypatch.setattr(sys, "platform", "linux")
+    monkeypatch.setattr(sys, "argv", ["auditwheel"])
+
+    # WHEN
+    retval = main()
+
+    # THEN
+    assert retval is None
+    captured = capsys.readouterr()
+    assert "usage: auditwheel [-h] [-V] [-v] command ..." in captured.out
+
--- a/tests/unit/test_main.py
+++ b/tests/unit/test_main.py
@@ -1,7 +1,11 @@
+import pytest
 import sys
 
 from auditwheel.main import main
 
+on_supported_platform = pytest.mark.skipif(
+    sys.platform != 'linux', reason="requires Linux system"
+)
 
 def test_unsupported_platform(monkeypatch):
     # GIVEN
@@ -14,9 +18,9 @@ def test_unsupported_platform(monkeypatch):
     assert retval == 1
 
 
+@on_supported_platform
 def test_help(monkeypatch, capsys):
     # GIVEN
-    monkeypatch.setattr(sys, "platform", "linux")
     monkeypatch.setattr(sys, "argv", ["auditwheel"])
 
     # WHEN


tests/unit/test_main.py
```diff
@@ -0,0 +1,70 @@
...
+    # GIVEN
+    monkeypatch.setattr(sys, "platform", "darwin")
+
+    # WHEN
+    retval = main()
+
+    # THEN
+    assert retval == 1
+    captured = capsys.readouterr()
+    assert captured.out == "Error: This tool only supports Linux\n"
```
takluyver: What's the goal of a test like this? You're testing that the code is exactly what's written. If we somehow add support for running on Darwin, or if we switch to a different way of checking the platform instead of `sys.platform`, we just have some extra work to change or discard the test as well.

Rather than trying to execute every line of code in the tests, I'd suggest we focus on the predictive value of the tests. If a test fails, what's the chance that it's identifying a real bug in the code under test? What's the chance we need to change the test to match changes in the code? If the second is much more likely than the first (like in this test, as far as I can see), it's a waste of time.

If it's important to get "100% coverage", I'd prefer to [exclude sections from coverage](https://coverage.readthedocs.io/en/v4.5.x/excluding.html) instead of writing hyper-specific tests just to make those lines run.

tests/unit/test_main.py
```diff
@@ -0,0 +1,70 @@
...
+    args.verbose = 1
+    args.func.return_value = 0
+    parser_mock.return_value.parse_args.return_value = args
+
+    # WHEN
+    retval = main()
+
+    # THEN
+    assert retval == 0
+    basic_cfg_mock.assert_called_once_with(level=logging.DEBUG)
```
takluyver: Again, how are these going to fail? If we change the argument parsing or the logging setup, we'd just have to change the tests to match. I find it hard to imagine any situation where these tests would fail and we'd change the `main()` function instead of the test.

tests/unit/test_main.py
lkollar: As we've agreed previously on not making it possible to run `auditwheel` on non-Linux platforms, I think it makes sense to have a test for it.

Would you be OK with having only one assert for the return value and no validation on the error message? At that point all we do is to set the platform to `linux` and make sure the tool errors out. I think this is better than simply excluding the logic from coverage and I don't see any other way to validate this.

tests/unit/test_main.py
lkollar: While I agree that the code tested here is unlikely to change without the test needing to change, if you simply want to make sure that the `-v` flag works correctly, how else do you implement the test for it? 

If we say that we don't test verbose mode because it would couple the test with the implementation, I would still opt to have the test, just so that it's validated. I think the only other option is to check that text the logger outputs includes `DEBUG`, but that would have to make assumptions on other parts of the system, which is worse in my opinion.

tests/unit/test_main.py
takluyver: I think it's better without checking the exact error message. It still seems a fairly pointless test, though: all the situations where I can imagine it failing would involve us fixing the test. It's really hard to imagine introducing a bug in the code which this would pick up.

I guess I don't think 100% test coverage is a useful goal. I like coverage information as a way of spotting chunks of code which aren't tested, but I think it's then a judgment call if you want to test them, and for things like this it's fine not to.

tests/unit/test_main.py
takluyver: I probably just wouldn't write a test for it. It works. There's no need to change it, and if we do switch to a new logging framework or something, the test will fail regardless of whether there's a bug.

If I really, really needed to test something like this, I'd try to find some operation which is guaranteed to produce debug level logging and, as you mention, verify that some of that is shown. That's ultimately what we care about. But I agree with you that this is rather an ugly test, and I probably wouldn't bother with it.

takluyver:
Sorry, I seem to be being very negative on your PRs.

Rather than me reviewing the details of specific tests, what I'd really like is to have a discussion with the other maintainers (@auvipy @mayeut) on what we want to do about tests for the project as a whole.

I'm arguing that aiming for 100% coverage is a waste of effort for all of us. We end up needing to write largely pointless tests just to get the coverage points, and changing anything involves extra work because you have to fix hyper-specific tests that assume the previous implementation. And then you have to add more tests for your new implementation to keep the needle at 100%. It's the programming equivalent of that myth about the man forever trying to push a rock up a hill, only for it to slip and roll down.

If I'm in the minority on this, and the others agree that we should have 100% coverage, I'll pipe down and let you get on with it.

lkollar:
@takluyver I completely agree that we should have a common understanding of how we implement the test suite. Given that we already have an issue for the unit test work, I have addressed your question there: #163.

tests/unit/test_main.py
```diff
@@ -0,0 +1,70 @@
+import logging
+import sys
+
+from unittest.mock import patch, Mock
+
+from auditwheel.main import main
+
+
+def test_non_linux_platform(monkeypatch, capsys):
```
mayeut: I'd change this by `test_unsupported_platform`

tests/unit/test_main.py
```diff
@@ -0,0 +1,70 @@
...
+    # GIVEN
+    monkeypatch.setattr(sys, "platform", "darwin")
```
mayeut: As mentioned by @takluyver, what if `darwin` is supported ? I'd say `not_a_platform`

tests/unit/test_main.py
mayeut: I would drop the string check as well.

tests/unit/test_main.py
mayeut: Same as @takluyver, I wouldn't bother with these 2 tests.

tests/unit/test_main.py
```diff
@@ -0,0 +1,70 @@
...
+    retval = main()
+
+    # THEN
+    assert retval == 0
+    basic_cfg_mock.assert_called_once_with(level=logging.DEBUG)
+
+
+def test_help(monkeypatch, capsys):
+    # GIVEN
+    monkeypatch.setattr(sys, "platform", "linux")
```
mayeut: Here, you're monkey patching the platform again. If the test if relevant (not sure but why not), then just skip the test on non linux platforms. This way, there's no correlation between the platform check and this test.

lkollar: This test checks that not providing any arguments will cause the help message to be displayed. The definition of a unit test is that it tests the SUT independent of its environment. If I let the platform check go through this would not be a real unit test as it would become environment dependent. 

Besides, I think it makes sense to validate that changes did not break the CLI logic on a non-Linux machine as well. I don't see why patching the platform is an issue here. Unless we stop supporting Linux, it's unlikely that this code will need to change.

lkollar:
I've made changes as requested by @takluyver and @mayeut, except for `test_help` which I would like to keep a unit test.

tests/unit/test_main.py
mayeut: I guess my point is that, if I ask someone to write a unit test for exactly what you said "This test checks that not providing any arguments will cause the help message to be displayed.", no one would ever monkeypatch the platform. That's reason enough to skip. Furthermore, if the check changes in CLI to use anything else than `sys.platform`, this test will fail and need change.

There are only 2 tests left, but before you forced push there would have been 3 places to change, maybe unlikely but still. I would at least define a fixture to keep this platform monkeypatch reusable.
I would skip it using:
```
skip_unsupported = pytest.mark.skipif(sys.platform != 'linux',
                                      reason="platform not supported")

@skip_unsupported
def test_help(monkeypatch, capsys):
```

or monkeypatch it:
```
@pytest.fixture
def _mock_platform_check(monkeypatch):
    monkeypatch.setattr(sys, "platform", "linux")


mock_platform_check = pytest.mark.usefixtures("_mock_platform_check")

@mock_platform_check
def test_help(monkeypatch, capsys):
```
that way, no need to modify everywhere in case of a change in CLI implementation.

The `skip_unsupported ` is at least not dependent on the usage of `sys.platform` in CLI but would fail to skip if the check in CLI changes to only allow x86/x86_64 architectures for example. Also, it'll still skip the tests if darwin/windows gets supported.

The `mock_platform_check` is still dependent on CLI check implementation, test will fail on other architectures if the check in CLI changes to only allow x86/x86_64 architectures for example.

A more robust way to mock or skip would be to have a function in the main module that does the platform check. It could be mocked or used for skipping in a more robust way.
```
def _platform_supported():
    if sys.platform != 'linux':
        print('Error: This tool only supports Linux')
        return False
    return True


def main():
    if not _platform_supported():
        return 1
```
This is probably taking a sledgehammer to crack a nut

tests/unit/test_main.py
mayeut: > I would drop the string check as well.

That wasn't clear at all.
While checking the content of the message is unnecessary, I think we should at least check there's something (we don't check it's meaningful but at least something was written out).
```
captured = capsys.readouterr()
assert len(captured.out) > 0
```

tests/unit/test_main.py
lkollar: > I guess my point is that, if I ask someone to write a unit test for exactly what you said "This test checks that not providing any arguments will cause the help message to be displayed.", no one would ever monkeypatch the platform.

I respectfully disagree. A unit test by definition is a test to validate functionality *in isolation*. If a unit test's result depends on what platform runs, it is not a unit test. Unit tests are supposed to test that the logic in the *unit under test* is correct and disregard all external factors (e.g. no external I/O or calls to anything which might do I/O should be allowed). The fact that often times such tests need to know implementation details in order to patch out external calls is inevitable and it becomes largely subjective to decide how many implementation details are we OK to accept in these tests. If the code is structured well, little mocking should be necessary, but this is not always the case and minimal changes to unit tests due to implementation changes are perfectly fine in my opinion.

lkollar:
I think the time we have spent so far arguing over `test_help` by far outweighs the time it would take to change the test whenever an implementation change breaks it. It also seems that there is no consensus among the maintainers whether this test is acceptable or not. At this point I can either close or land the PR but I do not want to spend more time on this change as I think my time can be much better used in other parts of the project. 

I will let the other maintainers decide what to do with this PR.

takluyver:
@lkollar - I'm sorry, I agree we've spent more time discussing this than we would ever have spent updating this test. But the discussion is really about how we approach testing as a whole, not this one specific test. If we can establish general principles for writing good tests, we'll ultimately save much more time across the codebase.

For `test_help`, I agree with @mayeut: just skip it on unsupported platforms. Our CI is running on Linux, so there's no point making other platforms pretend to be Linux for this.

This may be a personal preference, but I'd rather we didn't get too hung up on the definition of unit tests. It's definitely a useful concept to test smaller pieces of functionality separately, but I'd rather approach the separation pragmatically than strictly.

lkollar:
> For `test_help`, I agree with @mayeut: just skip it on unsupported platforms. Our CI is running on Linux, so there's no point making other platforms pretend to be Linux for this.

What about contributors who want to run the unit tests on non-Linux platforms? It seems totally arbitrary to me to decide to not patch here and let the test be environment specific in order to avoid refactoring the test in case the implementation changes, while anyone running the test suite on macOS breaking the test will not notice it until the CI fails. 

> This may be a personal preference, but I'd rather we didn't get too hung up on the definition of unit tests. It's definitely a useful concept to test smaller pieces of functionality separately, but I'd rather approach the separation pragmatically than strictly.

I actually think it's more pragmatic to stick to the original definition of a unit test and not let the environment affect the tests so that we can avoid subjective reasoning why a certain part should or should not be patched.

I think I complied with everything we have so far discussed except for skipping on non-Linux which I really don't want to do in a unit test. We seem to have two for and two against the PR in its current form, so I'll let someone else close or land it.

takluyver:
OK, I've switched it to skip on non-Linux platforms. I don't think it's terribly useful to develop on non-Linux platforms anyway, because you can't really verify the important functionality. And I prefer to avoid monkeypatching unless it's really necessary.

If the CI has passed when I next check (about to board a flight), I'll merge it.

PR #173, opened 2019-06-26
Clear rpath before setting

### LINKED ISSUE #29: Vendored library cannot be loaded

daa:
Due to bug in patchelf https://github.com/NixOS/patchelf/issues/94 if shared object has DT_RUNPATH attribute it is not replaced with DT_RPATH if --force-rpath is given, so one must clear all rpath attributes before setting new.
This pull request closes #29.
Also this pull request is resubmit of #133 because old showed too much changed files and it is easier for me to resubmit.

### DIFF
--- a/tests/integration/test_manylinux.py
+++ b/tests/integration/test_manylinux.py
@@ -351,7 +351,12 @@ def test_build_repair_pure_wheel(any_manylinux_container, io_folder):
 @pytest.mark.parametrize('dtag', ['rpath', 'runpath'])
 def test_build_wheel_depending_on_library_with_rpath(any_manylinux_container, docker_python,
                                                      io_folder, dtag):
-    # Test building a wheel that contains an extension depending on a library with RPATH set
+    # Test building a wheel that contains an extension depending on a library
+    # with RPATH or RUNPATH set.
+    # Following checks are performed:
+    # - check if RUNPATH is replaced by RPATH
+    # - check if RPATH location is correct, i.e. it is inside .libs directory
+    #   where all gathered libraries are put
 
     policy, manylinux_ctr = any_manylinux_container
 
--- a/tests/integration/test_manylinux.py
+++ b/tests/integration/test_manylinux.py
@@ -416,13 +416,8 @@ def test_build_wheel_depending_on_library_with_rpath(any_manylinux_container, do
                 with w.open(name) as f:
                     elf = ELFFile(io.BytesIO(f.read()))
                     dynamic = elf.get_section_by_name('.dynamic')
-                    rpath = runpath = None
-                    for t in dynamic.iter_tags():
-                        if t.entry.d_tag == 'DT_RPATH':
-                            rpath = t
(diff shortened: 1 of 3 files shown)


lkollar:
You probably want to squash your commits as there 15 in total and several have duplicate commit message. It doesn’t look quite right...

daa:
Squashed, sorry for inconvenience.

tests/integration/test_manylinux.py
```diff
@@ -406,7 +412,13 @@ def test_build_wheel_depending_on_library_with_rpath(any_manylinux_container, do
                 with w.open(name) as f:
                     elf = ELFFile(io.BytesIO(f.read()))
                     dynamic = elf.get_section_by_name('.dynamic')
+                    rpath = runpath = None
+                    for t in dynamic.iter_tags():
+                        if t.entry.d_tag == 'DT_RPATH':
+                            rpath = t
+                        elif t.entry.d_tag == 'DT_RUNPATH':
```
lkollar: Correct me if I'm wrong but I think the original logic could have left intact with the addition of an extra assertion:

`assert len([t for t in dynamic.iter_tags() if t.entry.d_tag == 'DT_RUNPATH']) == 0` instead of the ` assert runpath is None` line.

I think this is simpler and it also keeps the original assertion on having only one `DT_RPATH`, while the new change takes the last one in the list and doesn't check the length.

tests/integration/test_manylinux.py
```diff
@@ -348,8 +348,9 @@ def test_build_repair_pure_wheel(any_manylinux_container, io_folder):
     ]) in output.replace('\n', ' ')
 
 
+@pytest.mark.parametrize('dtag', ['rpath', 'runpath'])
 def test_build_wheel_depending_on_library_with_rpath(any_manylinux_container, docker_python,
-                                                     io_folder):
+                                                     io_folder, dtag):
     # Test building a wheel that contains an extension depending on a library with RPATH set
```
lkollar: Given that this is quite a complicated test I think it would be nice to have a few words on the additional checks on `RPATH`/`RUNPATH` you are adding now, to make it a bit easier to understand the code.

PR #178, opened 2019-07-09
Make integration tests faster

takluyver:
(Hopefully)

I noticed that all the tests which use docker repeat trying to pull the same images and do the same setup steps inside the containers. So this adds a layer of fixtures to create temporary docker images in session-scoped fixtures, deleting them when tests finish.

On my system, this makes the tests finish in under half the time they previously took (~800 seconds to ~320 seconds). It should also mean that all the docker based tests run in a consistent environment, even if an image is updated or a package released while the tests are running.

I've also told Travis to cache the wheels built in the tests. That won't affect the test time on this PR, because there's nothing cached yet, but once it's merged it should avoid the need to recompile numpy each time.

### DIFF
--- a/tests/integration/test_manylinux.py
+++ b/tests/integration/test_manylinux.py
@@ -59,8 +59,6 @@ def docker_start(image, volumes={}, env_variables={}):
     """
     # Make sure to use the latest public version of the docker image
     client = docker.from_env()
-    logger.info("Pulling docker image %r", image)
-    client.images.pull(image)
 
     dvolumes = {host: {'bind': ctr, 'mode': 'rw'}
                 for (ctr, host) in volumes.items()}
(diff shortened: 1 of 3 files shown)


takluyver:
On Travis, we've gone from 11-12 minutes for each test run to ~9 minutes. So not such a big saving, but still worth having. It also decreases the cost of adding more integration tests.

There should be a bit more speedup once it is using the cached numpy wheels.

tests/integration/test_manylinux.py
```diff
@@ -105,26 +107,69 @@ def io_folder(tmp_path):
...
+
+    Pulls the *base* image, runs *commands* inside it with *setup_env*, and
+    commits the result as a new image. The image is removed on exiting the
+    context.
+
+    Making temporary images like this avoids each test having to re-run the
+    same container setup steps.
+    """
+    client = docker.from_env()
+    logger.info("Pulling docker image %r", base)
```
lkollar: I wonder if we could just skip the pull step altogether. As far as I know the Docker daemon will pull the image anyway if it's not found locally, so in Travis we should always automatically get the latest version. For local testing this would mean faster startup if the image is already available locally. Even if the fixture is scoped to the session, repeatedly re-running a single test still incurs the cost of the pull every time the test is executed which adds up.

This would also enable testing with different versions of the image if needed, as various versions can be tagged locally for the test to pick up.

takluyver: I think that makes sense. I'll also add a note about it in the README.

takluyver:
Now with the cached numpy wheels we're down to about 5 minutes for each job. :-)

PR #182, opened 2019-07-18
Remove libcrypt.so.1 from library whitelist

takluyver:
Following https://github.com/pypa/manylinux/issues/305 : Fedora 30 no longer has `libcrypt.so.1` (it has `.2` instead), so IIUC it needs to be removed from the whitelist.

CC @njsmith .


njsmith:
This looks fine to me. The thing where there are multiple independent copies of the whitelist seems like it could probably be improved by some refactoring, but that's orthogonal....

Do we need to sort out the manylinux build images part before merging, or what's the plan there?

takluyver:
I'm not sure. Presumably with this change but without changing the build images, wheels that use `libcrypt` would end up bundling the `libcrypt.so.1` built from glibc. I don't have a good understanding of the possible consequences of that.

takluyver:
The PRs for the manylinux images were merged, so as far as I know this can go in (& into a release) at any point now.

Issue #185, opened 2019-08-06
Add manylinux2014 policy


lkollar:
As per https://www.python.org/dev/peps/pep-0599/.

di:
https://github.com/pypa/auditwheel/pull/192 has been merged, should this be closed?

mayeut:
cross referencing #202 which needs to be merged

dralley:
That issue is merged as well, now.

PR #192, opened 2019-09-22
manylinux2014 policy

mayeut:
Add support for manylinux2014 policy
Symbol versions were updated using latest centos:7 docker image by running the updated `scripts/calculate_symbol_versions.py`

c.f. #185
c.f. pypa/manylinux#338

tests in #193 (I'd rather keep them separated while waiting for official pip and manylinux image)


mayeut:
@lkollar, 

since you reviewed #141, can you please also review this one to get things going for `manylinux2014`

Thanks

auditwheel/policy/policy.json
```diff
@@ -40,6 +40,32 @@
...
+    {"name": "manylinux2014",
+     "priority": 80,
+     "symbol_versions": {
+         "GLIBC": ["2.0", "2.1", "2.1.1", "2.1.2", "2.1.3", "2.2", "2.2.1",
+                   "2.2.2", "2.2.3", "2.2.4", "2.2.5", "2.2.6", "2.3", "2.3.2",
+                   "2.3.3", "2.3.4", "2.4", "2.5", "2.6", "2.7", "2.8", "2.9",
+                   "2.10", "2.11", "2.12", "2.13", "2.14", "2.15", "2.16",
+                   "2.17"],
+         "CXXABI": ["1.3", "1.3.1", "1.3.2", "1.3.3", "1.3.4", "1.3.5",
+                    "1.3.6", "1.3.7", "TM_1"],
```
lkollar: What's the `TM_1` version? Seems odd to be a symbol version. I would take it out unless there is a good reason to keep it.

auditwheel/policy/policy.json
```diff
@@ -40,6 +40,32 @@
...
+         "GLIBCXX": ["3.4", "3.4.1", "3.4.2", "3.4.3", "3.4.4", "3.4.5",
+                     "3.4.6", "3.4.7", "3.4.8", "3.4.9", "3.4.10", "3.4.11",
+                     "3.4.12", "3.4.13", "3.4.14", "3.4.15", "3.4.16",
+                     "3.4.17", "3.4.18", "3.4.19"],
+         "GCC": ["3.0", "3.3", "3.3.1", "3.4", "3.4.2", "3.4.4", "4.0.0",
+                 "4.2.0", "4.3.0", "4.7.0", "4.8.0"]
```
lkollar: PEP 599 goes up to `GCC_4.8.5`. Should we add that too?

auditwheel/policy/policy.json
mayeut: as mentioned in description, all symbols version come from `scripts/calculate_symbol_versions.py`. I've not checked which symbol gets versioned with it.
This version is also present on all distros I checked for compatibility: https://github.com/pypa/manylinux/pull/304#issuecomment-498042781

auditwheel/policy/policy.json
mayeut: I think that PEP shall be updated. I'll open a PR.
here are the results for i686/aarch64 architectures (note the presence of `TM_1` here too):
i686:
```
{
"GLIBC": ["2.0", "2.1", "2.1.1", "2.1.2", "2.1.3", "2.2", "2.2.1", "2.2.2", "2.2.3", "2.2.4", "2.2.6", "2.3", "2.3.2", "2.3.3", "2.3.4", "2.4", "2.5", "2.6", "2.7", "2.8", "2.9", "2.10", "2.11", "2.12", "2.13", "2.14", "2.15", "2.16", "2.17"],
"CXXABI": ["1.3", "1.3.1", "1.3.2", "1.3.3", "1.3.4", "1.3.5", "1.3.6", "1.3.7", "TM_1"],
"GLIBCXX": ["3.4", "3.4.1", "3.4.2", "3.4.3", "3.4.4", "3.4.5", "3.4.6", "3.4.7", "3.4.8", "3.4.9", "3.4.10", "3.4.11", "3.4.12", "3.4.13", "3.4.14", "3.4.15", "3.4.16", "3.4.17", "3.4.18", "3.4.19"],
"GCC": ["3.0", "3.3", "3.3.1", "3.4", "3.4.2", "4.0.0", "4.2.0", "4.3.0", "4.4.0", "4.5.0", "4.7.0", "4.8.0"]
}
```
aarch64:
```
{
"GLIBC": ["2.0", "2.17", "2.18"],
"CXXABI": ["1.3", "1.3.1", "1.3.2", "1.3.3", "1.3.4", "1.3.5", "1.3.6", "1.3.7", "TM_1"],
"GLIBCXX": ["3.4", "3.4.1", "3.4.2", "3.4.3", "3.4.4", "3.4.5", "3.4.6", "3.4.7", "3.4.8", "3.4.9", "3.4.10", "3.4.11", "3.4.12", "3.4.13", "3.4.14", "3.4.15", "3.4.16", "3.4.17", "3.4.18", "3.4.19"],
"GCC": ["3.0", "3.3", "3.3.1", "3.4", "3.4.2", "3.4.4", "4.0.0", "4.2.0", "4.3.0", "4.5.0", "4.7.0"]
}
```

di: Good catch, that's a typo. GCC 4.8.5 uses the `GCC_4.8.0` symbol version. I've made the change in https://github.com/python/peps/pull/1181.

auditwheel/policy/policy.json
di: Did you need to modify `calculate_symbol_versions.py` to get it to output this version? Looks like it expects all versions to be integers separated by periods:

https://github.com/pypa/auditwheel/blob/3191806fa535c118d4b521ff701e522e9fb639e4/scripts/calculate_symbol_versions.py#L44

Also note that this script is already doing some filtering of the versions:

https://github.com/pypa/auditwheel/blob/3191806fa535c118d4b521ff701e522e9fb639e4/scripts/calculate_symbol_versions.py#L67

So maybe it does make sense to filter `TM_1` here. (I have no idea what it is.)

auditwheel/policy/__init__.py
```diff
@@ -17,8 +17,10 @@
 _PLATFORM_REPLACEMENT_MAP = {
     'manylinux1_x86_64': ['linux_x86_64'],
     'manylinux2010_x86_64': ['linux_x86_64'],
+    'manylinux2014_x86_64': ['linux_x86_64'],
     'manylinux1_i686': ['linux_i686'],
     'manylinux2010_i686': ['linux_i686'],
+    'manylinux2014_i686': ['linux_i686'],
```
di: Similar to https://github.com/pypa/warehouse/pull/6684, we also need to add the other architectures that PEP 599 supports here.

auditwheel/policy/policy.json
mayeut: > Did you need to modify calculate_symbol_versions.py to get it to output this version?

Yes, first commit of this PR

> Also note that this script is already doing some filtering of the versions

It makes sense to filter `PRIVATE` symbols which I guess are used between glibc libraries for internal usage. `TM_1` provides __cxa_tm_cleanup function which is a support function used by the compiler given what I understand from https://gcc.gnu.org/onlinedocs/gcc-7.1.0/libitm/The-libitm-ABI.html. I don't think it makes sense to filter it.

auditwheel/policy/__init__.py
mayeut: I'd like to get this done in a second time. There's been literally no tests on those other platforms so unless we say that support for other architectures is experimental for now, I'd rather do it in 2 times.

auditwheel/policy/policy.json
di: > Yes, first commit of this PR

Apologies, I missed this.

takluyver: If we're going to allow the `TM_1` symbol version, presumably the PEP should be updated to mention that?

lkollar: Ping @mayeut. I think this is the last comment which needs addressing. We can then land this PR.

mayeut: Right, I will post a message to libstdc++ mailing list over the weekend to get their input on this symbol version.

mayeut: https://gcc.gnu.org/ml/libstdc++/2019-10/msg00023.html

takluyver: The reply indicates that it can be relied upon, although there's incredulity that a Python extension would ever use it. We're not using it, though, just allowing libraries that may use it. 

So unless there's any reason auditwheel should block libraries from using transactional memory extensions, we can leave it in here and add it to the PEP.

mayeut: Opened a [PEP Pull Request](https://github.com/python/peps/pull/1201)

di:
https://github.com/python/peps/pull/1201 has been merged, is there anything left to do here?

PR #193, opened 2019-09-25
WIP - manylinux2014 policy tests

mayeut:
Add tests on top of #192
pip built from https://github.com/mayeut/pip/tree/manylinux2014
docker image is automatically built from branch used for pypa/manylinux#348

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,14 @@
+os: linux
 language: python
-dist: xenial
+
 matrix:
   include:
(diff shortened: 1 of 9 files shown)


lkollar:
So it seems this has to blockers: the new `manylinux2014` image and an updated `pip` which supports the new tag. I assume you plan to wait for these before going ahead with the PR?

mayeut:
Tests are now built on top of #202

PR #199, opened 2019-10-03
Run manylinux tests using current python version

mayeut:
Run integration tests on manylinux using the python version used to test auditwheel.
Since travis-ci is running 3.5, 3.6 & 3.7, we're better off testing those versions inside manylinux containers rather than testing 3 times the exact same thing.

### DIFF
--- a/tests/integration/test_manylinux.py
+++ b/tests/integration/test_manylinux.py
@@ -6,6 +6,7 @@ import io
 import os
 import os.path as op
 import shutil
+import sys
 import logging
 import zipfile
 from auditwheel.policy import get_priority_by_name
@@ -23,13 +24,15 @@ MANYLINUX_IMAGES = {
     'manylinux2010': MANYLINUX2010_IMAGE_ID,
 }
 DOCKER_CONTAINER_NAME = 'auditwheel-test-manylinux'
-PYTHON_IMAGE_ID = 'python:3.5'
+PYTHON_MAJ_MIN = [str(i) for i in sys.version_info[:2]]
+PYTHON_ABI = 'cp{0}-cp{0}m'.format(''.join(PYTHON_MAJ_MIN))
+PYTHON_IMAGE_ID = 'python:' + '.'.join(PYTHON_MAJ_MIN)
 DEVTOOLSET = {
     'manylinux1': 'devtoolset-2',
     'manylinux2010': 'devtoolset-8',
 }
 PATH_DIRS = [
-    '/opt/python/cp35-cp35m/bin',
+    '/opt/python/{}/bin'.format(PYTHON_ABI),
     '/opt/rh/{devtoolset}/root/usr/bin',
     '/usr/local/sbin',
     '/usr/local/bin',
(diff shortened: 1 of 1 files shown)


tests/integration/test_manylinux.py
```diff
@@ -23,13 +24,15 @@
...
+PYTHON_MAJ_MIN = [str(i) for i in sys.version_info[:2]]
+PYTHON_ABI = 'cp{0}-cp{0}m'.format(''.join(PYTHON_MAJ_MIN))
+PYTHON_IMAGE_ID = 'python:' + '.'.join(PYTHON_MAJ_MIN)
 DEVTOOLSET = {
     'manylinux1': 'devtoolset-2',
     'manylinux2010': 'devtoolset-8',
 }
 PATH_DIRS = [
-    '/opt/python/cp35-cp35m/bin',
+    '/opt/python/{}/bin'.format(PYTHON_ABI),
```
lkollar: If you want to make this dynamic, you could just do `os.path.dirname(sys.executable)`.

tests/integration/test_manylinux.py
```diff
@@ -302,7 +305,7 @@ def test_build_wheel_with_image_dependencies(with_dependency, any_manylinux_cont
     assert 'manylinux' not in orig_wheel
 
     repair_command = \
-        'LD_LIBRARY_PATH=/auditwheel_src/tests/integration/testdependencies '\
+        'LD_LIBRARY_PATH=/auditwheel_src/tests/integration/testdependencies:$LD_LIBRARY_PATH '\
```
lkollar: This change seems unrelated. It's probably harmless, but I wonder why you are changing it.

tests/integration/test_manylinux.py
```diff
@@ -436,21 +439,21 @@ def test_build_wheel_depending_on_library_with_rpath(any_manylinux_container, do
     ).format(policy=policy, orig_wheel=orig_wheel)
     docker_exec(
         manylinux_ctr,
-        ['bash', '-c', 'LD_LIBRARY_PATH=/auditwheel_src/tests/integration/testrpath/a ' + repair_command],
+        ['bash', '-c', 'LD_LIBRARY_PATH=/auditwheel_src/tests/integration/testrpath/a:$LD_LIBRARY_PATH ' + repair_command],
```
lkollar: See comment above.

tests/integration/test_manylinux.py
```diff
@@ -23,13 +24,15 @@
     'manylinux2010': MANYLINUX2010_IMAGE_ID,
 }
 DOCKER_CONTAINER_NAME = 'auditwheel-test-manylinux'
-PYTHON_IMAGE_ID = 'python:3.5'
+PYTHON_MAJ_MIN = [str(i) for i in sys.version_info[:2]]
```
lkollar: Another approach to this would be to use the `wheel` package which we already depend on. That should provide a better guess at the ABI tag as it handles things like figuring out the allocator, etc.
```
from wheel.pep425tags import get_abbr_impl, get_impl_ver, get_abi_tag
PYTHON_ABI = "{0}{1}-{2}".format(get_abbr_impl(), get_impl_ver(), get_abi_tag()))
```

tests/integration/test_manylinux.py
mayeut: When running python 3.7 tests, `libcrypto.so.2` needs to be found. Overwriting `LD_LIBRARY_PATH` not preserving the existing prevents `libcrypto.so.2` to be found.

tests/integration/test_manylinux.py
mayeut: idem

tests/integration/test_manylinux.py
mayeut: We can't if I'm not mistaken. `sys.executable` relates to the `python` interpreter running the tests while `/opt/python/{}/bin'.format(PYTHON_ABI)` relates to the `python` interpreter running in the container.

tests/integration/test_manylinux.py
mayeut: > We can't if I'm not mistaken. sys.executable relates to the python interpreter running the tests while /opt/python/{}/bin'.format(PYTHON_ABI) relates to the python interpreter running in the `manylinux` container.

This is also valid here I think but if there's something better, please correct me.

PR #203, opened 2019-10-17
Advertise python 3.8 support and run python 3.8 in CI

mayeut:


setup.cfg
```diff
@@ -16,6 +16,7 @@ classifier =
         Programming Language :: Python :: 3.5
         Programming Language :: Python :: 3.6
         Programming Language :: Python :: 3.7
+        Programming Language :: Python :: 3.8
```
takluyver: I tend to leave off the classifiers for minor Python versions to avoid the need for this kind of change, and use the `python_requires` metadata to specify a minimum supported version. This is a matter of preference, though.

mayeut: IMHO, it needs changing every 12 months or so, it's not such a hassle and advertise that it's tested on python 3.8 as well. The `python_requires` should always be used and that's not the case for `auditwheel`. I pushed another commit fixing this.

tests/integration/test_manylinux.py
```diff
@@ -25,7 +25,7 @@
 }
 DOCKER_CONTAINER_NAME = 'auditwheel-test-manylinux'
 PYTHON_MAJ_MIN = [str(i) for i in sys.version_info[:2]]
-PYTHON_ABI = 'cp{0}-cp{0}m'.format(''.join(PYTHON_MAJ_MIN))
+PYTHON_ABI = 'cp{0}-cp{0}{1}'.format(''.join(PYTHON_MAJ_MIN), 'm' if int(PYTHON_MAJ_MIN[1]) < 8 else '')
```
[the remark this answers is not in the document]
mayeut: fixed

Issue #212, opened 2019-12-12
auditwheel repair marked internal so files as shared library dependencies


wdirons:
Thanks to everyone that worked on adding manylinux 2014 support to auditwheel.

I'm trying to get build a manylinux 2014 whl file of TensorFlow for ppc64le, using the docker image `quay.io/pypa/manylinux2014_ppc64le`

This image contains auditwheel 3.0.0.0rc1

auditwheel show says I'm close to being compliant, I just have one external dependency to resolve:

```
# auditwheel show tf_nightly-2.1.0.dev20191210-cp36-cp36m-linux_ppc64le.whl

tf_nightly-2.1.0.dev20191210-cp36-cp36m-linux_ppc64le.whl is
consistent with the following platform tag: "linux_ppc64le".

The wheel references external versioned symbols in these system-
provided shared libraries: libc.so.6 with versions {'GLIBC_2.17'},
libpthread.so.0 with versions {'GLIBC_2.17'}, libgcc_s.so.1 with
versions {'GCC_4.2.0', 'GCC_3.0', 'GCC_3.3', 'GLIBC_2.0'}, libm.so.6
with versions {'GLIBC_2.17'}, ld64.so.2 with versions {'GLIBC_2.17'},
libdl.so.2 with versions {'GLIBC_2.17'}, librt.so.1 with versions
{'GLIBC_2.17'}, _pywrap_tensorflow_internal.so with versions
{'tensorflow'}

The following external shared libraries are required by the wheel:
{
    "ld64.so.2": "/lib64/ld-2.17.so",
    "libc.so.6": "/lib64/libc-2.17.so",
    "libdl.so.2": "/lib64/libdl-2.17.so",
    "libgcc_s.so.1": "/lib64/libgcc_s-4.8.5-20150702.so.1",
    "libm.so.6": "/lib64/libm-2.17.so",
    "libpthread.so.0": "/lib64/libpthread-2.17.so",
    "librt.so.1": "/lib64/librt-2.17.so"
}

In order to achieve the tag platform tag "manylinux2014_ppc64le" the
following shared library dependencies will need to be eliminated:

ld64.so.2
```

auditwheel repair appears to resolve the issue:
```
# auditwheel repair tf_nightly-2.1.0.dev20191210-cp36-cp36m-linux_ppc64le.whl
INFO:auditwheel.main_repair:Repairing tf_nightly-2.1.0.dev20191210-cp36-cp36m-linux_ppc64le.whl
[...]

PR #213, opened 2019-12-24
Always exclude ELF dynamic linker/loader

mayeut:
`s390x`/`ppc64le` ELF dynamic linkers do not start with `ld-linux`and therefore, were not excluded from analysis.

Relates to #212 
Fixes https://github.com/pypa/manylinux/issues/412

The tests did not reveal this issue and shall be updated. I don't know how to get this done since all symbols in the linker are private (or seem to be).

@wdirons, @st-pasha, any simple test we could include to reproduce this behavior in tests ?

### DIFF
--- a/auditwheel/elfutils.py
+++ b/auditwheel/elfutils.py
@@ -45,7 +45,8 @@ def elf_find_versioned_symbols(elf: ELFFile) -> Iterator[Tuple[str, str]]:
 
     if section is not None:
         for verneed, verneed_iter in section.iter_versions():
-            if verneed.name.startswith('ld-linux'):
+            if verneed.name.startswith('ld-linux') or \
+                    verneed.name in ['ld64.so.2', 'ld64.so.1']:
                 continue
             for vernaux in verneed_iter:
                 yield (verneed.name,
--- a/auditwheel/policy/external_references.py
+++ b/auditwheel/policy/external_references.py
@@ -16,8 +16,11 @@ def lddtree_external_references(lddtree: Dict, wheel_path: str):
 
     def filter_libs(libs, whitelist):
         for lib in libs:
-            if 'ld-linux' in lib:
-                # always exclude ld-linux.so
+            if 'ld-linux' in lib or lib in ['ld64.so.2', 'ld64.so.1']:
+                # always exclude ELF dynamic linker/loader
+                # 'ld64.so.2' on s390x
+                # 'ld64.so.1' on ppc64le
+                # 'ld-linux*' on other platforms
                 continue
             if LIBPYTHON_RE.match(lib):
                 # always exclude libpythonXY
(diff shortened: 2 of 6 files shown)


di:
I added two integration tests for this using numpy wheels for s390x and ppc64le thanks to @wdirons in #212 -- if these aren't what we were looking for, please feel free to amend/remove my commit.

Also, I think the s390x build will probably fail here due to #228, but we should probably merge this anyways.

tests/integration/test_elf_dynamic_linkers.py
```diff
@@ -0,0 +1,18 @@
...
+    winfo = analyze_wheel_abi(
+        'tests/integration/numpy-1.18.1-cp36-cp36m-linux_ppc64le.whl')
+    external_libs = winfo.external_refs['manylinux2014_ppc64le']['libs']
+    assert len(external_libs) == 0
+
+@pytest.mark.skipif(platform.machine() != 's390x', reason='only supported on s390x')
+def test_analyze_wheel_abi_elf_dynamic_linkers_s390x():
+    winfo = analyze_wheel_abi(
+        'tests/integration/numpy-1.18.1-cp36-cp36m-linux_s390x.whl')
+    external_libs = winfo.external_refs['manylinux2014_ppc64le']['libs']
```
wdirons: Is this a cut and paste mistake? Mention of ppc64le in an s390x testcase?

di: Yep, thanks!

di:
Tests are now passing, what do you think @mayeut?

mayeut:
@wdirons,  @di, 
thanks for the test case.
I'd really like not to have 25MB+ stored in the repo. I'll try to see if I can get this shrunk a bit.

wdirons:
I'm trying to figure out how to modify the test_elfutils.py test for this instance.

I have noticed, the existing testcase for ld-linux is not testing what it should as this change still passes the test:

```
diff --git a/tests/unit/test_elfutils.py b/tests/unit/test_elfutils.py
index 23e31e7..045ad5b 100644
--- a/tests/unit/test_elfutils.py
+++ b/tests/unit/test_elfutils.py
@@ -99,7 +99,7 @@ class TestElfFindVersionedSymbols:
         # GIVEN
         elf = Mock()
         verneed = Mock()
-        verneed.configure_mock(name="ld-linux")
+        verneed.configure_mock(name="foo-bar")
         elf.get_section_by_name.return_value.iter_versions.return_value = (
             (verneed, []),
         )
```

mayeut:
I finally managed to get some time to find a small test case.
Relying on TLS creates a dependency on the ELF dynamic linker/loader.

Failures can be seen here when the fix is not here: https://travis-ci.com/github/mayeut/auditwheel/jobs/320711188 & https://travis-ci.com/github/mayeut/auditwheel/jobs/320711187

mayeut:
CI not passing but passing here https://travis-ci.com/github/mayeut/auditwheel/builds/160833993
so going ahead with merge

PR #223, opened 2020-01-28
fixes for auditwheel to depend on wheel 0.34

Czaki:
wheel in version 0.34 introduce fix for pypy tags https://github.com/pypa/wheel/issues/328

### DIFF
--- a/auditwheel/wheeltools.py
+++ b/auditwheel/wheeltools.py
@@ -6,15 +6,16 @@ Tools that aren't specific to delocation
 import os
 from os.path import (join as pjoin, abspath, relpath, exists, sep as psep,
                      splitext, dirname, basename)
+import sys
 import glob
 import hashlib
 import csv
 from itertools import product
 import logging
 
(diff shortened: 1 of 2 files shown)


auditwheel/wheeltools.py
```diff
@@ -27,6 +28,16 @@ class WheelToolsError(Exception):
     pass
 
 
+def open_for_csv(name, mode):
```
lkollar: Given that auditwheel only runs on Python 3 and there is only one place where this function is called, we could safely replace that call with `open(name, mode, newline='', encoding='utf-8')`. That avoids the unnecessary version check and makes the code more readable.

Czaki: I introduce this change. Could you restart failed steep. It fail on timeout error.

lkollar: Restarted the ppc64 build.

Czaki:
@lkollar  When you think it will be next release?

Issue #218, opened 2020-01-21
Auditwheel addtag returns stack trace when given a none-any wheel


jvanstraten:
Steps to reproduce:

```text
# Download any none-any wheel (just chose one randomly here)
$ curl -L https://files.pythonhosted.org/packages/a7/1e/2d18651bb48933b6089833df8d1e44182b2899784e26072f94860440ae8c/plumbum-1.6.8-py2.py3-none-any.whl > plumbum-1.6.8-py2.py3-none-any.whl
# Run auditwheel, receive stack trace
$ auditwheel addtag plumbum-1.6.8-py2.py3-none-any.whl
INFO:auditwheel.main_addtag:plumbum-1.6.8-py2.py3-none-any.whl receives the following tag: "manylinux1_x86_64".
INFO:auditwheel.main_addtag:Use ``auditwheel show`` for more details
INFO:auditwheel.wheeltools:Previous filename tags: any
Traceback (most recent call last):
  File "/usr/local/bin/auditwheel", line 10, in <module>
    sys.exit(main())
  File "/opt/_internal/cpython-3.7.5/lib/python3.7/site-packages/auditwheel/main.py", line 50, in main
    rval = args.func(args, p)
  File "/opt/_internal/cpython-3.7.5/lib/python3.7/site-packages/auditwheel/main_addtag.py", line 48, in execute
    out_wheel = add_platforms(ctx, [wheel_abi.overall_tag])
  File "/opt/_internal/cpython-3.7.5/lib/python3.7/site-packages/auditwheel/wheeltools.py", line 215, in add_platforms
    remove_platforms.append('any')
AttributeError: 'tuple' object has no attribute 'append'
```

[The offending line](https://github.com/pypa/auditwheel/blame/d5cd0f670eecf9a94be67911e18d3c2dea8ffc55/auditwheel/wheeltools.py#L215) calls `append()` on something that, according to the function's docs, can be any iterable, and by default is `()`. That obviously doesn't work. I fixed this locally by adding the obvious

```python
remove_platforms = list(remove_platforms)
```

which seems to work, but honestly I don't know much about how wheels work internally, so I'm not sure what I'm doing. That's why we're using manylinux/auditwheel, after all. I do understand that running auditwheel on a none-any wheel is not usually a sane thing to do in the first place, but at the very least I wouldn't expect a stack trace.

This was run on the manylinux 2010 container with Python 3.6, but I'm fairly certain it doesn't matter. According to the blame that code has been there for a while.

PR #224, opened 2020-01-28
Correctly detect non-platform wheels

lkollar:
We are not checking if there are any ELF files in the wheel when
inspecting the wheel ABI. This can lead to misleading results when
invoking 'show' and an uncaught exception in other modes.

Resolves: #218


lkollar:
This is ready for review. The s390x build is still failing due to running out of space on the image: https://travis-ci.community/t/no-space-left-on-device-for-system-z/5954/9.

PR #226, opened 2020-01-30
Fix obsolete wheel usage in addtag

lkollar:
We upgraded to wheel 0.34.1 and the API is slightly different. The
addtag command was not updated for the change.

The included test fails without the fix.

### DIFF
--- a/auditwheel/main_addtag.py
+++ b/auditwheel/main_addtag.py
@@ -22,13 +22,13 @@ def configure_parser(sub_parsers):
 
 def execute(args, p):
     import os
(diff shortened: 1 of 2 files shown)


lkollar:
The s390x build is still failing due to #228 and it's unclear when Travis will fix this. Meanwhile the addtag command is broken so I think it's best if we land this without the s390x build.

di:
Looks like it's actually failing due to https://github.com/pypa/manylinux/issues/473. Agreed that this can land w/o the s309x build passing.

PR #237, opened 2020-04-28
Ensure that system-copied lib is writable before running patchelf

rmcgibbo:
It appears that at least some versions of patchelf will fail to patch files that are marked as read-only (source: https://github.com/NixOS/nixpkgs/issues/14440). We hit this bug when trying to run auditwheel repair on a system in which the external libraries copied in by `repair` are read-only.

This patch adds the writable bit after copying them into the wheel, before running auditwheel.

cc @jgullingsrud, who ID'd this.


rmcgibbo:
I feel a bit shamed by the coverage report. Should I make a test with some non-writable system packages that need to me chmoded?

auvipy:
it would be great if you can update tests :)

lkollar:
I think unit testing it might be a bit involved as the function does quite a few things. An integration test is probably simpler, provided it runs inside Docker and makes sure to reset the file permission back to the original avoid polluting other tests (a fixture could be ideal for this). One of the existing manylinux tests could be modified for this.

rmcgibbo:
Alright, I added a test. I confirmed locally that without the patch to copylib, the test does fail with an error from patchelf.

tests/unit/test_repair.py
```diff
@@ -0,0 +1,41 @@
...
+from auditwheel.repair import copylib
+from auditwheel.tmpdirs import InTemporaryDirectory
+from auditwheel.lddtree import lddtree
+
+@pytest.fixture
+def systemlib():
+    for libdata in lddtree(sys.executable)["libs"].values():
+        # return one of the arbitrary libraries that python is linked
+        # against, like libc.so.6. any one will do.         
+        return libdata["realpath"]
```
lkollar, line 15: You can omit the for loop:
```suggestion
    # return one of the arbitrary libraries that python is linked
    # against, like libc.so.6. any one will do.         
    lib, *_ = lddtree(sys.executable)["libs"].values()
    return lib["realpath"]
```

tests/unit/test_repair.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+on_supported_platform = pytest.mark.skipif(
+    sys.platform != 'linux', reason="requires Linux system"
+)
+
+@on_supported_platform
```
lkollar, line 22: Given these two tests end up invoking `patchelf` I think they belong in the integration tests.

rmcgibbo: Okay, done. I'm not sure if the integration tests are run in an environment with patchelf installed, or if I need to move these two tests under docker as well.

rmcgibbo:
How do we run python integration tests that require patchelf? The docker stuff in test_manylinux.py seems best for running bash and subprocesses, rather than than simple python tests of a single function in auditwheel.

rmcgibbo:
I think this is probably best to merge as-is.

PR #248, opened 2020-06-18
Add mypy to lint checks

lkollar:


lkollar:
The s390x build is failing due to some issue with Travis.

Issue #281, opened 2021-01-19
Add flag to prevent multiple wheel repairs


ofek:
https://github.com/pypa/auditwheel/blob/7502b9dd8b308083ec47918f2ebf87e419838393/auditwheel/main_repair.py#L97-L109

from CI logs:

```
+ auditwheel repair wheelhouse/coincurve-14.0.0-cp36-cp36m-linux_i686.whl --plat manylinux2014_i686 -w out
INFO:auditwheel.main_repair:Repairing coincurve-14.0.0-cp36-cp36m-linux_i686.whl
INFO:auditwheel.wheeltools:Previous filename tags: linux_i686
INFO:auditwheel.wheeltools:New filename tags: manylinux2014_i686
INFO:auditwheel.wheeltools:Previous WHEEL info tags: cp36-cp36m-linux_i686
INFO:auditwheel.wheeltools:New WHEEL info tags: cp36-cp36m-manylinux2014_i686
INFO:auditwheel.main_repair:Wheel is eligible for a higher priority tag. You requested manylinux2014_i686 but I have found this wheel is eligible for manylinux2010_i686.
INFO:auditwheel.wheeltools:Previous filename tags: linux_i686
INFO:auditwheel.wheeltools:New filename tags: manylinux2010_i686
INFO:auditwheel.wheeltools:Previous WHEEL info tags: cp36-cp36m-linux_i686
INFO:auditwheel.wheeltools:New WHEEL info tags: cp36-cp36m-manylinux2010_i686
INFO:auditwheel.main_repair:
Fixed-up wheel written to /out/coincurve-14.0.0-cp36-cp36m-manylinux2010_i686.whl
+ cp out/coincurve-14.0.0-cp36-cp36m-manylinux2010_i686.whl out/coincurve-14.0.0-cp36-cp36m-manylinux2014_i686.whl /io/dist
```

Ideally, only one wheel would be produced: the one with the requested platform or the one with the higher priority tag.

If one doesn't add an additional step to delete one then it's just wasted space on PyPI currently.

lkollar:
If I understand correctly the problem here is that you requested manylinux2014 but auditwheel generated manylinux2010 tag? The wheel itself is manylinux2010 compliant, which means it will work on both manylinux2010 and manylinux2014 platforms and uploading the 2010 version should be sufficient. What is the advantage in producing a 2014 variant as well?

ofek:
@lkollar It produces the 2014 one I requested _and_ unrequested 2010

[...]

PR #283, opened 2021-02-06
Fix: Patch RPATHs of non-Python extension dependencies

### LINKED ISSUE #136: Non-python extension dependencies with external dependencies are not recursively patched

bdice:
This PR resolves #136, and replaces the open (stalled) PR #139 from @thomaslima (closes #139).

## Summary of the problem
From my comment in #136, slightly edited for clarity:
> We have a few Python extensions that link to `libfreud.so` and `libtbb.so`, neither of which are Python extensions. `libfreud.so` is also linked to `libtbb.so`. auditwheel correctly relocates the RPATHs of our Python extensions, but does not alter the RPATH of `libfreud.so` to point to the relocated `libtbb.so`.

## Solution
Add non-Python extensions to the dictionary of `full_external_refs` in `get_wheel_elfdata` in `wheel_abi.py`.

## Context
This PR adds the fix and ports over tests from a previous PR #134 that was reverted in #190 due to tests failing (I think the failures were caused by changes in the testing infrastructure, but I'm not 100% sure). Also related to this is PR #245 which handles a related issue that @thomaslima worked on in #140.

## Testing
The `hello` package in `tests/integration/nonpy_rpath` is a minimal-ish example of such a package with a non-Python extension library with an RPATH that needs to be repaired. The tests I added are based on @thomaslima's previous work, updated to match newer conventions in the testing code for auditwheel. I think the tests are relatively complete. **I manually verified that the tests fail without the fix.**


auditwheel/wheel_abi.py
```diff
@@ -118,10 +118,16 @@ def get_wheel_elfdata(wheel_fn: str):
...
                 full_elftree[fn] = nonpy_elftree[fn]
-                full_external_refs[fn] = lddtree_external_references(
-                    nonpy_elftree[fn], ctx.path)
 
-    log.debug(json.dumps(full_elftree, indent=4))
+            # Even if a non-pyextension ELF file is not needed, we
+            # should include it as an external references, because
+            # they might also require external libraries.
+            full_external_refs[fn] = lddtree_external_references(
+                nonpy_elftree[fn], ctx.path)
```
bdice, line 126: The core of the fix is just to move this line out of the "if" statement above.

thomaslima:
Thank you so much for providing this. I really wasn't able to work on the original PRs. I wonder if you have used any of the code that I wrote for this.

bdice:
> Thank you so much for providing this. I really wasn't able to work on the original PRs. I wonder if you have used any of the code that I wrote for this.

Yes, I based this on your previous PRs. You can see in the commit history that I cherry-picked your previous work onto a new branch before updating the code, so that you receive credit in the git history. 👍

mayeut:
Thanks for working on this @bdice, @thomaslima 
I will review this next week-end at the latest.

mayeut (changes requested):
The functional code change is OK however, I find the test case a bit too complicated and might conflict (or more exactly, the test would become useless) with #152 

A simpler, more contained test should work as well here, maybe something like `tests/integration/testdependencies` but going one level deeper in dependencies to cover the issue would work if I understand correctly ?

bdice:
@mayeut I agree that your suggestion for a simpler test extension would be good. I had hoped it would be good enough as-is, but you're right that won't work any more with #152. It will take me some time to make these changes -- I will try to have a simpler test extension by the end of the week. Is there a release schedule or other timeline to be aware of? I would like to get this fix into the next release if possible.

psavery:
@bdice Thanks for working on this. I can confirm this fixes #136 for us.

Until it's in a release, we'll patch this change into the docker image we're using with cibuildwheel.

bdice:
Update: I just finished writing a new, simpler test case. I have to do some local testing and code cleanup before re-requesting review. I anticipate having updated tests for this PR finished in the next 2 days.

mayeut:
Thanks for the update @bdice 
> Is there a release schedule or other timeline to be aware of? I would like to get this fix into the next release if possible.

We have no release schedule for `auditwheel`. Given this fixes an issue, I'd plan to release `auditwheel` at most 1 week after merging the PR depending on my free time to do so. The other PRs are not blocker for a release.

bdice:
@mayeut I tried to create a new test, in `tests/integration/test_nonpy_dependencies`, but it's not catching the problem correctly (the tests pass without the fix). I had to learn the Python C API and a lot of gcc linker behaviors that I haven't dealt with before, and it looks like I didn't quite get the "failing case" correct. Maybe I can use the previous test infrastructure but just use another external library that isn't `zlib`? Would that fix the conflict with #152 and be alright with you as a maintainer? Do you have a recommended library that I should try to use? I appreciate the help.

mayeut:
@bdice,
Anything other than `zlib` and `expat` would be fine I think. You can probably try using symbols from `libxcrypt`, it won't be whitelisted again and can be grafted properly.

> Is it expected that auditwheel will use the oldest possible platform tag for the ABIs in the wheel, even if --plat is given a newer tag?

Yes, see also #281

PR #284, opened 2021-02-10
Only create a single repaired wheel

lkollar:
When repairing wheels, the result can be compliant with a lower-level
manylinux spec and in this case auditwheel creates another .whl file for
the lower spec as well. This is unnecessary as the lower spec can always
be used on newer compliant platforms. By only producing a single wheel,
we can probably save some space on PyPI as automated build scripts won't
upload multiple copies of the same wheel.

This also avoids an extra call to repair_wheel which can speed things up
for larger libraries.

Closes: #281

### DIFF
--- a/auditwheel/main_repair.py
+++ b/auditwheel/main_repair.py
@@ -1,10 +1,13 @@
+import logging
+import os
 from os.path import isfile, exists, abspath, basename
 
-from auditwheel.patcher import Patchelf
+from .patcher import Patchelf
 from .policy import (load_policies, get_policy_name, get_priority_by_name,
                      POLICY_PRIORITY_HIGHEST)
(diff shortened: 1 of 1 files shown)


lkollar:
I haven't added a flag for this because I don't think there is any point in retaining the previous behaviour, but I might be missing something. Anybody see any problems with this?

mayeut:
@lkollar,

I will try to get a little history on the why because I know this has been reported and discussed before.
I fully agree with the fact that only one wheel should be produced however, I do find the information of the policy used to repair the wheel useful and I think there might be a way not to lose it.
I will get back here with my findings over the week-end.

mayeut:
@lkollar,

I left some comment in #281

PR #287, opened 2021-02-18
Ensure policies in `policy.json` are compliant with PEP600

### LINKED ISSUE #286: ncurses 5 -> 6 transition means we probably need to drop some libraries from the manylinux whitelist
The [PEP 513 whitelist](https://www.python.org/dev/peps/pep-0513/#the-manylinux1-policy) includes `libpanelw.so.5` and `libncursesw.so.5`, both of which are part of ncurses version 5.

But apparently the new hostness is [ncurses 6, which breaks ABI with ncurses 5](https://www.gnu.org/software/ncurses/). I'm told that Fedora 24 and successors have already switched to ncurses 6, with the ncurses 5 libraries being available in a compatibility package, but not installed by default. So... I guess these have to be dropped from the whitelist.

mayeut:
This also removes 2 non existing symbols from manylinux1 i686 policy and removes ncurses librairies from manylinux1 whitelist.

Now that [PEP600](https://www.python.org/dev/peps/pep-0600/) supersedes previous manylinux PEPs, it should be enforced:
> When this PEP is accepted, the previous manylinux PEPs will receive a final update noting that they are no longer maintained and referring to this PEP.

Fix #286


mayeut:
Failure on TravisCI ppc64le is unrelated to the PR.

auditwheel/policy/__init__.py
```diff
@@ -23,9 +24,30 @@ def get_arch_name() -> str:
...
 
 
+def _validate_pep600_compliance(policies) -> None:
+    symbol_versions: Dict[str, Dict[str, Set[str]]] = {}
+    lib_whitelist: Set[str] = set()
+    for policy in sorted(policies, key=lambda x: x['priority'], reverse=True):
+        if policy['name'] == 'linux':
+            continue
+        if not lib_whitelist.issubset(set(policy['lib_whitelist'])):
+            raise ValueError('Invalid "policy.json" file')
```
lkollar, line 34: How about making the string more explicit on what's wrong with the policy check? That could make it easier to diagnose the issue when editing the JSON.

lkollar (commented):
Looks good, just one suggestion.

auditwheel/policy/__init__.py
mayeut: I added more information and tests for that. Is that better now ?

lkollar: Great, thanks!

lkollar (approved):

PR #298, opened 2021-03-06
Fix: Patch RPATHs of non-Python extension dependencies

### LINKED ISSUE #136: Non-python extension dependencies with external dependencies are not recursively patched

mayeut:
This PR resolves #136, and replaces the open (stalled) PR #139 from @thomaslima (closes #139) and rebase/rework the open PR #283 from @bdice (closes #283)

Summary in #283 is the nice one, so go there for more information.

Tests are much simpler here and still reproduce the issue without the fix of course.

Thanks @bdice, @thomaslima

### DIFF
--- a/auditwheel/wheel_abi.py
+++ b/auditwheel/wheel_abi.py
@@ -118,10 +118,16 @@ def get_wheel_elfdata(wheel_fn: str):
             # we should walk its elftree.
             if basename(fn) not in needed_libs:
                 full_elftree[fn] = nonpy_elftree[fn]
-                full_external_refs[fn] = lddtree_external_references(
-                    nonpy_elftree[fn], ctx.path)
 
-    log.debug(json.dumps(full_elftree, indent=4))
+            # Even if a non-pyextension ELF file is not needed, we
+            # should include it as an external references, because
+            # they might also require external libraries.
+            full_external_refs[fn] = lddtree_external_references(
+                nonpy_elftree[fn], ctx.path)
+
+    log.debug('full_elftree:\n%s', json.dumps(full_elftree, indent=4))
+    log.debug('full_external_refs (will be repaired):\n%s',
+              json.dumps(full_external_refs, indent=4))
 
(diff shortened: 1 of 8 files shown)


bdice:
@mayeut Thank you so much for helping with this, I really appreciate it. I got a little stuck on #283 — some pieces of the code needed for testing this fix are new to me, but I’ve learned a lot! I just read over the PR and had a couple minor suggestions that I will mark in a review.

auditwheel/wheel_abi.py
```diff
@@ -118,10 +118,16 @@ def get_wheel_elfdata(wheel_fn: str):
             # we should walk its elftree.
             if basename(fn) not in needed_libs:
                 full_elftree[fn] = nonpy_elftree[fn]
-                full_external_refs[fn] = lddtree_external_references(
-                    nonpy_elftree[fn], ctx.path)
 
-    log.debug(json.dumps(full_elftree, indent=4))
+            # Even if a non-pyextension ELF file is not needed, we
+            # should include it as an external references, because
+            # they might also require external libraries.
```
bdice, line 124: This might help clarify.
```suggestion
            # Even if a non-pyextension ELF file is not needed, we
            # should include it as an external reference, because
            # it might require additional external libraries.
```

bdice (approved):
This looks good to me. This PR is much cleaner than what I was doing on #283. Nice work, @mayeut! 👍

bdice:
I'm going to use this PR in my wheel building process until it's merged. Here's a simple script for anyone else that needs it:

```bash
git clone https://github.com/mayeut/auditwheel.git -b issue136 auditwheel
cd auditwheel
pip install .
```

For some reason, the version appears to be `auditwheel 1.8.1.dev251` instead of something like 3.3.1, but the software works and I verified that the bug fix works as intended.

auvipy (approved):

PR #306, opened 2021-05-11
Fix undeclared dependency on setuptools

DavHau:
auditwheel without setuptools raises:
`ModuleNotFoundError: No module named 'pkg_resources'`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,5 +4,6 @@ from setuptools import setup
 
 setup(
     setup_requires=['pbr'],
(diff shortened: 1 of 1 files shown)


lkollar:
Thanks for the PR @DavHau. Since the project uses PBR, the dependencies should be listed in requirements.txt. I've submitted another PR which fixes this and switches to `importlib.metadata`.

DavHau:
Thanks

PR #307, opened 2021-05-21
Remove dependency on `pkg_resources`

lkollar:
The only reason we needed this was to look at the version metadata and
the installation location. We can do this with `importlib.metadata`
which is part of Python since 3.8 and is significantly faster to import
than `pkg_resources`.

Closes: #306


mayeut (approved):
LGTM

PR #308, opened 2021-05-22
Use python slim images to run tests

mayeut:
This should speed-up the tests a bit.


mayeut:
@lkollar,

This was mostly to check that TravisCI builds were still triggered following an issue on the Travis CI pypa account.
Nevertheless, I still think it's a good thing to use the slim image.

auvipy (approved):
slim is better

lkollar (approved):

Issue #241, opened 2020-05-19
Do not place OpenGL libraries into wheels on manylinux*


mathstuf:
These libraries really need to come from the end machine, not the wheel. They don't work without a backing video driver and may not work if the package was built on a machine with a different driver than the end machine (e.g., mixing an nvidia driver `libOpenGL` with a Mesa DRI `libOpenGL`). The backing DRI libraries aren't packaged anyways, so they're of limited utility either way.

mattip:
I see that `libGL` is already part of the [allowlist policies](https://github.com/pypa/auditwheel/blob/main/src/auditwheel/policy/manylinux-policy.json#L66). Is there also a `libOpenGL` that should be allowlist-ed?

mathstuf:
Note that the PR that closed this only adds an `--exclude` flag. Is there some way to have PyPI add this flag when verifying uploads? Or does this need a formal PEP?

Issue #250, opened 2020-06-24
How to make manylinux package with MKL dependency


TonyXiang8787:
I am working on a Python package with C++ extensions which have Intel MKL dependency. The extension module links to `libmkl_rt.so`. I have put `mkl>=2019` in the requirements in `setup.py`.

The MKL itself is a `manylinux1` package in the PyPI:
https://pypi.org/project/mkl/#files

However, if I run `auditwheel repair` to produce a wheel file with `manylinux2014` tag. It will copy the `libmkl_rt.so` into the wheel file. This is of course not needed (and also not preferred). The requirement on MKL is already specified and can be found in the PyPI library. And the `auditwheel repair` only copies the `libmkl_rt.so` which is merely an entry for the real MKL library.

Is it a intended behavior of `auditwheel repair` to copy all the external dynamic dependencies into the wheel file, even if the dependency is a  `manylinux1` package in the PyPI and the requirements are specified in `setup.py`? If yes, how I can avoid that the `libmkl_rt.so`  when producing a `manylinux2014` wheel?

fzimmermann89:
Does anybody have an advice how to handle this?

mayeut:
> Is it a intended behavior of auditwheel repair to copy all the external dynamic dependencies into the wheel file

Yes, it's the intended behavior of auditwheel. It grafts all external dynamic dependencies into the wheel file (with the exception of those listed in the [whitelist](https://github.com/pypa/auditwheel/blob/4b05be803c46619c1de329233c45d625380598b3/auditwheel/policy/policy.json#L51-L60)) for it to be installable on any manylinux compliant system.

> even if the dependency is a manylinux1 package in the PyPI and the requirements are specified in setup.py?

That's an undefined behavior. PyPI is only there to manage python dependencies, not native ones and thus, auditwheel does not rely on such information. The fact that Intel uses PyPI to redistribute libraries is either for convenience or for some legal reason (e.g. the license forbids redistribution or modification).

**auditwheel will modify libraries when they're grafted in**

> If yes, how I can avoid that the libmkl_rt.so when producing a manylinux2014 wheel?

You can remove the library from the needed libraries and have a mechanism to dynamically load `libmkl_rt.so` before you import your extension. That's [the way onnxruntime does it](https://github.com/microsoft/onnxruntime/blob/017b94ab366eafead3a94b10b8510dc372212ae7/setup.py#L131-L149) for CUDA.

TonyXiang8787:
[...]

Issue #161, opened 2019-05-14
libz.so.1.2.3 on CentOS 6 causes repair to segfault


romaingz:
Hi,

OS: **CentOS 6.10**
libz.so: **1.2.3**
python: **3.6.6**
patchelf: **9.0**

I've just run into an issue with `libz.so` with its `1.2.3` version on **CentOS 6**.
When repairing I get the following error:
```bash
DEBUG:auditwheel.repair:Grafting: /lib64/libz.so.1.2.3 -> .libs_MyLib/libz-70dc5473.so.1.2.3
Traceback (most recent call last):
  File "/home/path/to/venv/bin/auditwheel", line 11, in <module>
    sys.exit(main())
  File "/home/path/to/venv/lib/python3.6/site-packages/auditwheel/main.py", line 47, in main
    rval = args.func(args, p)
  File "/home/path/to/venv/lib/python3.6/site-packages/auditwheel/main_repair.py", line 79, in execute
    update_tags=args.UPDATE_TAGS)
  File "/home/path/to/venv/lib/python3.6/site-packages/auditwheel/repair.py", line 88, in repair_wheel
    new_soname, new_path = copylib(src_path, dest_dir)
  File "/home/path/to/venv/lib/python3.6/site-packages/auditwheel/repair.py", line 141, in copylib
    check_call(['patchelf', '--set-soname', new_soname, dest_path])
  File "/home/other/path/to/Python/3.6.6/lib/python3.6/subprocess.py", line 291, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['patchelf', '--set-soname', 'libz-70dc5473.so.1.2.3', '.libs_MyLib/libz-70dc5473.so.1.2.3']' died with <Signals.SIGSEGV: 11>
```
I also get a `segfault` if I simply try to `--print-soname`.
Note that there is no similar issue with the `libz.1.2.7` on CentOS 7.

It turns out that I really don't need `libz` to be included in this process, since (at least on all our CentOS systems) it comes already installed.

I temporarily worked around that by wrapping the `copylib` into a `try/except` block:
```python
try:
    new_soname, new_path = copylib(src_path, dest_dir)
    soname_map[soname] = (new_soname, new_path)
    check_call(['patchelf', '--replace-needed', soname, new_soname, fn])
except Exception as e:
    logger.warning(str(e))
```
and it works.
[...]

Issue #152, opened 2019-04-24
We should probably allow libz linkage, and maybe libexpat


njsmith:
Here's the output of `ldd python3` on my system (Ubuntu 18.10):

```
~$ ldd /usr/bin/python3
	linux-vdso.so.1 (0x00007ffd9ddbb000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f1f0e73b000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f1f0e71a000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f1f0e714000)
	libutil.so.1 => /lib/x86_64-linux-gnu/libutil.so.1 (0x00007f1f0e70f000)
	libexpat.so.1 => /lib/x86_64-linux-gnu/libexpat.so.1 (0x00007f1f0e6d2000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f1f0e4b5000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f1f0e326000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f1f0e94c000)
```

Most of that is glibc or kernel stuff. The exceptions are `libexpat.so.1`, and `libz.so.1`.

`ldd /usr/bin/python2` gives a similar list, except it's missing `libexpat.so.1`.

The standard auditwheel profiles don't allow wheels to link against `libz.so.1` or `libexpat.so.1` – they get vendored instead. (See also: https://github.com/pypa/manylinux/issues/296)

However... this doesn't actually work!

Because of how ELF symbol resolution works, it doesn't matter whether you vendor these libraries: if they show up in `ldd <the interpreter>`, then any symbols in the system version of those libraries *will* be used by all python extensions, no matter what they vendor or link against. Basically the system version of `libz.so.1` and `libexpat.so.1` are getting `LD_PRELOAD`ed into every extension module.

And apparently... this works? I haven't seen any bugs about it. And it's plausible to me that these libraries have extremely stable ABIs.

So maybe we should add them to the list of libraries that auditwheel will allow you to depend on without vendoring... the vendoring isn't doing much anyway.

mayeut:
The dump for python interpreter on manylinux1 and manylinux2010 images are not quit the same.
[...]

PR #310, opened 2021-05-26
Add --include and --exclude options to auditwheel repair

rossant:
This is a quick fix to solve an issue I have with my project and for which auditwheel seemed to help. I don't know if this approach is suitable to other users of auditwheel.

I develop a C graphics library ([datoviz](https://github.com/datoviz/datoviz/)) that depends on Vulkan and comes with Cython bindings. [I'm trying to build a pip wheel](https://github.com/datoviz/datoviz/issues/13) for Linux (at least Ubuntu 20.04, that would be a start).

1. As a first step, I compile the C project and get a `libdatoviz.so` shared library. 
2. Then, I build the Cython library, which has a dynamic dependency to libdatoviz.
3. I'd like to bundle both the Cython extension module, and libdatoviz, in the same wheel.
4. I tried to use `auditwheel repair`. It included dozens of other dependencies in the wheel, including libvulkan and many graphics-related dependent libraries. The compiled wheel installs properly, but it doesn't work properly (there are issues related to windowing, GPU access, etc). Related: https://github.com/pypa/auditwheel/issues/241
5. I made some changes to auditwheel (this pull request) to exclude libvulkan and a few other libraries. Some issues were fixed, but not others.
6. As another approach, I tried to bundle *just* libdatoviz, and no other library in the wheel. That seems to work, at least on my computer. Chances are that the wheel may not work on other Linux-based operating systems though.

Any help would be appreciated, regarding either this pull request, and/or a way to solve the issue I'm facing. Thanks!

### DIFF
--- a/auditwheel/main_repair.py
+++ b/auditwheel/main_repair.py
@@ -65,6 +65,12 @@ below.
                    action='store_true',
                    help='Strip symbols in the resulting wheel',
                    default=False)
+    p.add_argument('--include',
+                   dest='INCLUDE',
+                   help='Only include these libraries')
+    p.add_argument('--exclude',
+                   dest='EXCLUDE',
+                   help='Exclude these libraries')
     p.add_argument('--only-plat',
                    dest='ONLY_PLAT',
                    action='store_true',
@@ -126,7 +132,10 @@ def execute(args, p):
                              out_dir=args.WHEEL_DIR,
                              update_tags=args.UPDATE_TAGS,
                              patcher=patcher,
-                             strip=args.STRIP)
+                             strip=args.STRIP,
+                             include=(args.INCLUDE or '').split(','),
+                             exclude=(args.EXCLUDE or '').split(','),
+                             )
 
     if out_wheel is not None:
         logger.info('\nFixed-up wheel written to %s', out_wheel)
--- a/auditwheel/repair.py
+++ b/auditwheel/repair.py
@@ -34,8 +34,8 @@ def _is_in_list(soname: str, items: Optional[List[str]]) -> str:
             return item
 
 
-def _filter(l: List[str]) -> List[str]:
-    return [_.strip() for _ in l if _.strip()]
+def _filter(items: List[str]) -> List[str]:
+    return [_.strip() for _ in items if _.strip()]
 
 
 def repair_wheel(wheel_path: str, abis: List[str], lib_sdir: str, out_dir: str,
--- a/auditwheel/repair.py
+++ b/auditwheel/repair.py
@@ -28,14 +28,15 @@ WHEEL_INFO_RE = re.compile(
     re.VERBOSE).match
 
 
-def _is_in_list(soname: str, items: Optional[List[str]]) -> str:
-    for item in items:
+def _is_in_list(soname: str, items: Optional[List[str]]) -> Optional[str]:
+    for item in (items or []):
         if item in soname:
             return item
+    return None
 
 
-def _filter(items: List[str]) -> List[str]:
-    return [_.strip() for _ in items if _.strip()]
+def _filter(items: Optional[List[str]]) -> List[str]:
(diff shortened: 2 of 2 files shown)


mayeut:
Thanks for your contribution @rossant 

Having an `--exclude` option, while dangerous, allows to work around those kind of issues in auditwheel (e.g. libvulkan but probably also `libOpenCL`, `libOpenVX`...)

Fixing those kind of issues requires a considerable amount of time related to the analysis of what's available on which distro, will it be compatible, is there any kind of commitment from those libraries on ABI stability for the foreseeable future...

This would allow projects using this option to do the analysis for their specific need. If the analysis is complete and allows the libraries to be whitelisted per PEP 600, then it might be better for the community to get auditwheel to update its policies.

I'd rather not have the `--include` option present but only the `--exclude` one at it is explicit what's not being included.

Having the `--exclude` option would help mitigate #241, #250, #161 and #152 while waiting for a more generic "foolproof" fix.

@lkollar, any thoughts ?

mayeut:
@rossant,
If you expect your wheel to only be ubuntu 20.04 compliant, then it's not a good idea to upload it to PyPI. It will only confuse end-users to see a manylinux wheel that doesn't work and will end up with reports to your repo tracker and/or [PyPI issue tracker](https://github.com/pypa/warehouse/issues).
In this case, your better of publishing your specific wheel as a release artifact on GitHub for example and update your docs to mention how to install it.
see also https://github.com/pypa/warehouse/issues/5420

rossant:
@mayeut makes sense, thanks. Are there guidelines somewhere about how to properly test a manylinux wheel on a sufficiently wide range of distributions to ensure it reaches the bar for PyPI?

mayeut:
> Are there guidelines somewhere about how to properly test a manylinux wheel on a sufficiently wide range of distributions to ensure it reaches the bar for PyPI?

There's no such guidelines to my knowledge. For simple packages, running tests in the manylinux docker is almost always enough. I'd say (personal opinion) add to that a recent distro if you have "simple" external dependencies. When you have not so simple dependencies it's a bit more complex I guess.

[PEP 600 only state (short extract, I advise to look at full spec)](https://www.python.org/dev/peps/pep-0600/#core-definition):
 > the wheel's creator promises that the wheel will work on any mainstream Linux distro that uses glibc version ${GLIBCMAJOR}.${GLIBCMINOR} or later
...
> The word "mainstream" is intentionally somewhat vague, and should be interpreted expansively. The goal is to rule out weird homebrew Linux systems; generally any distro you've actually heard of should be considered "mainstream".

For a `manylinux_2_31` wheel (ubuntu 20.04 uses glibc 2.31), given the dependencies you want to exclude, I'd look at the non EOLed distros in https://github.com/mayeut/pep600_compliance#distro-compatibility starting at `manylinux_2_31`.
In theory, if one of those distros can't run your wheel, it should not be eligible for manylinux distribution (I would add an exception on photon which is a headless only distro, given you're using the GPU, this distro probably does not apply).

martinRenou:
> Having an --exclude option, while dangerous, allows to work around those kind of issues in auditwheel (e.g. libvulkan but probably also libOpenCL, libOpenVX...)

I'd also be interested in this change being included!

@rossant are you planning on updating your PR? If time is not on your side I can open another PR myself adding only the `--exclude` option as suggested by @mayeut (while giving you credit for the commit)

rossant:
> @rossant are you planning on updating your PR? If time is not on your side I can open another PR myself adding only the `--exclude` option as suggested by @mayeut (while giving you credit for the commit)

I'd definitely appreciate some help! I won't have time to update my PR in the short term

martinRenou:
Ok :) I'll push a PR soonish

martinRenou:
See https://github.com/pypa/auditwheel/pull/368

mayeut:
closing in favor of #368

Issue #305, opened 2021-04-12
PEP 656 musllinux support


uranusjr:
[PEP 656](https://www.python.org/dev/peps/pep-0656/) proposed `musllinux`, a counterpart of `manylinux` for Linux distros running on [musl libc](https://musl.libc.org/). I’ve recently implemented support for installing `musllinux` wheels (pypa/packaging#411), and are now looking for a place to implement wheel generation logic.

Do you think auditwheel is a good place for this? Many things are the same auditing manylinux and musllinux wheels (the Linux part), but from my (very little) understanding, many parts in auditwheel assume glibc and would need some refactoring to work with musl, so there’s case to create a new tool for musllinux wheel auditing as well. My main end goal is to add musl to the support matrix of cibuildwheel (joerick/cibuildwheel#627), so either approach would ultimately be fine.

Edit from maintainers:
List of tasks that would need to be done in order to add support for musllinux.
This list is subject to changes and will be updated with feedback from comments.

- [x] Rename manylinux policy `auditwheel/policy/policy.json` -> `auditwheel/policy/manylinux-policy.json` #311 
- [x] Prepare a musllinux policy `auditwheel/policy/musllinux-policy.json` following the same schema. #315 
- [x] Detect if running a musl or glibc linux #315 
- [x] Select the correct policy file based on previous detection #315
- [x] Filter musllinux policies to only retain the one matching the system's musl version #315 https://github.com/pypa/auditwheel/issues/305#issuecomment-835840263
- [x] Fix ldtree to behave properly on musl #315 

This seems to be what's needed for now.
Be warned, [most of the work is probably `musllinux-policy.json`](https://github.com/pypa/auditwheel/issues/305#issuecomment-841493587).

mayeut:
Hi @uranusjr,

Sorry it took this long for someone to answer.
 
I think it makes sense to add musllinux support to auditwheel.
As you said, many things should be the same between musllinux & manylinux. Of course, there are also differences for glibc vs musl that need to be taken care of but I wouldn't say that auditwheel assumes glibc in its code.

One a of the major difference I foresee is that, because glibc uses symbol versioning, everything is done through the policy wether we're talking about glibc libraries, gcc/g++ libraries. Symbol versions are used to compute the effective policy and thus wether or not it's compatible with the requested one.
[...]

PR #311, opened 2021-05-27
refactor: rename policy.json to manylinux-policy.json

mayeut:
This prepares for other policies to be added with meaningful names.
Contributes to #305 resolution where `musllinux-policy.json` would be added.


mayeut:
Simple enough to require a review.
CI passing is enough.

Issue #258, opened 2020-08-15
Replace `unzip` usage with Python's `zipfile`


FFY00:
Would it be possible to replace the `unzip` usage with Python's `zipfile`? I see no reason to have an external dependency such as this.

https://github.com/pypa/auditwheel/blob/8fc14ac8943fd1cd8f6b350ce6f0432f749ef17c/auditwheel/tools.py#L43

lkollar:
According to the comment [above](https://github.com/pypa/auditwheel/blob/8fc14ac8943fd1cd8f6b350ce6f0432f749ef17c/auditwheel/tools.py#L41-L42), `unzip` is used to work around an issue where permissions are not correctly preserved. If we can ensure the behaviour is preserved, we can replace it with `ZipFile`, but it won't be a trivial change probably.

Alternatively, once https://github.com/pypa/wheel/pull/357 is landed and `wheel` will have a public API, we might be able to use that and drop the code dealing with ZIP files.

FFY00:
Sure, although you can still access the permissions in `zipfile.ZipInfo` IIRC. You can use `infolist` to set the permissions after the extraction, should be a pretty straight forward patch.

The reason I am asking this is because `unzip` is a custom external program. It is not part of any operating system. Some Linux distributions might include it by default, but definitely not all, it might also be missing.

PR #312, opened 2021-07-02
Extract using zipfile preserving permission

kumaraditya303:
Fixes https://github.com/pypa/auditwheel/issues/258

### DIFF
--- a/auditwheel/tools.py
+++ b/auditwheel/tools.py
@@ -1,11 +1,11 @@
 import argparse
 import os
 import shutil
+import subprocess
+import zipfile
 from glob import glob
 from os.path import join as pjoin
 from typing import Any, Iterable, List
-import zipfile
-import subprocess
 
 
 def unique_by_index(sequence: Iterable[Any]) -> List[Any]:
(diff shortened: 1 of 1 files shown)


kumaraditya303:
@mayeut Can you approve running workflows ?

auvipy (changes requested):
can you add test too

kumaraditya303:
> can you add test too

That would require adding a wheel file with permissions, should I add such a wheel file and then test its permission after extract ?

kumaraditya303:
@mayeut If you can add a test on the master branch then I can rebase on it and we can check if this change works as expected as the earlier was working.

mayeut:
@kumaraditya303,

Thanks for the PR.

> If you can add a test on the master branch then I can rebase on it and we can check if this change works as expected as the earlier was working.

I unfortunately won't have time for this in the foreseeable future. If you have time to add the test as @auvipy asked, I think it would be enough for now (or at least it would help doing some more checks based on this new test). 

> That would require adding a wheel file with permissions, should I add such a wheel file and then test its permission after extract ?

Yes, the smaller possible wheel file but covering as much permission bits as possible.

PR #313, opened 2021-07-05
Add support for musllinux

lkollar:
Closes: #305 

This is a work in progress.

PR #315, opened 2021-07-17
Add support for musllinux

mayeut:
This aims to modify less things than #313 at first to ease the review.
Once there are sufficient tests, the added complexity of #313 might be something to look into as it might be a good step forward to cross-repairing/cross-checking wheels.

@lkollar, thanks for the work in #313 (and in manylinux). What do you think of doing this in 2 steps as mentioned above?
1. add support for musllinux without extensively modifying the code base for manylinux
2. once we can test everything properly, refactor policy loading/management

### DIFF
--- a/auditwheel/lddtree.py
+++ b/auditwheel/lddtree.py
@@ -17,9 +17,12 @@ import glob
 import errno
 import logging
 import functools
+from pathlib import Path
 from typing import List, Dict, Optional, Any, Tuple
 
 from elftools.elf.elffile import ELFFile
+from .libc import get_libc, Libc
+
 
 log = logging.getLogger(__name__)
 __all__ = ['lddtree']
@@ -195,11 +198,32 @@ def load_ld_paths(root: str = '/', prefix: str = '') -> Dict[str, List[str]]:
             # on a per-ELF basis so it can get turned into the right thing.
             ldpaths['env'] = parse_ld_paths(env_ldpath, path='')
 
-    # Load up /etc/ld.so.conf.
-    ldpaths['conf'] = parse_ld_so_conf(root + prefix + '/etc/ld.so.conf',
-                                       root=root)
-    # the trusted directories are not necessarily in ld.so.conf
-    ldpaths['conf'].extend(['/lib', '/lib64/', '/usr/lib', '/usr/lib64'])
+    libc = get_libc()
+    if libc == Libc.MUSL:
+        # from https://git.musl-libc.org/cgit/musl/tree/ldso
+        # /dynlink.c?id=3f701faace7addc75d16dea8a6cd769fa5b3f260#n1063
+        root_prefix = Path(root) / prefix
+        ld_musl = list((root_prefix / 'etc').glob("ld-musl-*.path"))
+        assert len(ld_musl) <= 1
+        if len(ld_musl) == 0:
+            ldpaths['conf'] = [
+                root + '/lib',
+                root + '/usr/local/lib',
+                root + '/usr/lib'
+            ]
+        else:
+            ldpaths['conf'] = []
+            for ldpath in ld_musl[0].read_text().split(':'):
+                ldpath_stripped = ldpath.strip()
+                if ldpath_stripped == "":
+                    continue
+                ldpaths['conf'].append(root + ldpath_stripped)
+    else:
+        # Load up /etc/ld.so.conf.
+        ldpaths['conf'] = parse_ld_so_conf(root + prefix + '/etc/ld.so.conf',
+                                           root=root)
+        # the trusted directories are not necessarily in ld.so.conf
+        ldpaths['conf'].extend(['/lib', '/lib64/', '/usr/lib', '/usr/lib64'])
     log.debug('linker ldpaths: %s', ldpaths)
     return ldpaths
 
--- a/auditwheel/policy/musllinux-policy.json
+++ b/auditwheel/policy/musllinux-policy.json
@@ -1,0 +1,48 @@
+[
+  {"name": "linux",
+   "aliases": [],
+   "priority": 0,
+   "symbol_versions": {},
+   "lib_whitelist": []
+  },
+  {"name": "musllinux_1_1",
+    "aliases": [],
+    "priority": 100,
+    "symbol_versions": {
+      "i686": {
+      },
+      "x86_64": {
+      },
+      "aarch64": {
+      },
+      "ppc64le": {
+      },
+      "s390x": {
+      },
+      "armv7l": {
+      }
+    },
+    "lib_whitelist": [
+      "libc.so"
+    ]},
+    {"name": "musllinux_1_2",
+    "aliases": [],
+    "priority": 90,
+    "symbol_versions": {
+      "i686": {
+      },
+      "x86_64": {
+      },
+      "aarch64": {
+      },
+      "ppc64le": {
+      },
+      "s390x": {
+      },
+      "armv7l": {
+      }
+    },
+    "lib_whitelist": [
+      "libc.so"
+    ]}
+]
--- a/auditwheel/error.py
+++ b/auditwheel/error.py
@@ -1,0 +1,6 @@
+class AuditwheelException(Exception):
+    pass
+
+
+class InvalidLibc(AuditwheelException):
+    pass
--- a/auditwheel/libc.py
+++ b/auditwheel/libc.py
@@ -1,0 +1,23 @@
+import logging
+from enum import IntEnum
+
+from .error import InvalidLibc
+from .musllinux import find_musl_libc
+
+
+logger = logging.getLogger(__name__)
+
+
+class Libc(IntEnum):
+    GLIBC = 1,
+    MUSL = 2,
+
+
+def get_libc() -> Libc:
+    try:
+        find_musl_libc()
+        logger.debug("Detected musl libc")
+        return Libc.MUSL
+    except InvalidLibc:
+        logger.debug("Falling back to GNU libc")
+        return Libc.GLIBC
--- a/auditwheel/musllinux.py
+++ b/auditwheel/musllinux.py
@@ -1,0 +1,60 @@
+import logging
+import pathlib
+import subprocess
+import re
+from typing import NamedTuple
+
+from auditwheel.error import InvalidLibc
+
+LOG = logging.getLogger(__name__)
+
+
+class MuslVersion(NamedTuple):
+    major: int
+    minor: int
+    patch: int
+
+
+def find_musl_libc() -> pathlib.Path:
+    try:
+        ldd = subprocess.check_output(["ldd", "/bin/ls"], errors='strict')
+    except (subprocess.CalledProcessError, FileNotFoundError):
+        LOG.error("Failed to determine libc version", exc_info=True)
+        raise InvalidLibc
+
+    match = re.search(
+        r"libc\.musl-(?P<platform>\w+)\.so.1 "  # TODO drop the platform
+        r"=> (?P<path>[/\-\w.]+)",
+        ldd)
+
+    if not match:
+        raise InvalidLibc
+
+    return pathlib.Path(match.group("path"))
+
+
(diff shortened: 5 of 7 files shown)


tests/integration/test_manylinux.py
```diff
@@ -232,526 +186,608 @@ def assert_show_output(manylinux_ctr, wheel, expected_tag, strict):
...
-        docker_python,
-        ['python', '-c',
-         'from sys import exit; from testsimple import run; exit(run())']
-    )
+        # Repair the wheel using the appropriate manylinux container
+        repair_command = \
+            f'auditwheel repair --plat {policy} --only-plat -w /io /io/{orig_wheel}'
+        docker_exec(
+            manylinux_ctr,
+            ["bash", "-c", "LD_LIBRARY_PATH=/usr/local/lib " + repair_command]
```
mayeut, line 598: I had to add LD_LIBRARY_PATH for musllinux and I don't really know why yet. That's not needed on manylinux.

lkollar: What library was not found without LD_LIBRARY_PATH?

lkollar:
> @lkollar, thanks for the work in #313 (and in manylinux). What do you think of doing this in 2 steps as mentioned above?
> 
> 1. add support for musllinux without extensively modifying the code base for manylinux
> 2. once we can test everything properly, refactor policy loading/management

Sounds good to me. TBH I wasn't completely happy with how the policy loading ended up looking and I agree that it would be better to come back to this and properly refactor it later. I'll close my PR in favour of this one.

tests/integration/test_manylinux.py
mayeut: That would be `libcrypt.so.2`

mayeut: I guess it's related to `parse_ld_so_conf` in `lddtree.py`. There's something to be parsed on manylinux but not on musllinux. There might be something else to be done on musl...

auditwheel/policy/musllinux-policy.json
```diff
@@ -0,0 +1,48 @@
...
+      "x86_64": {
+      },
+      "aarch64": {
+      },
+      "ppc64le": {
+      },
+      "s390x": {
+      },
+      "armv7l": {
+      }
```
pvizeli, line 43:
```suggestion
      }
      "armv6l": {
      }
```

It's still happening that a small system with musl also runs armv6. Unlike manylinux, musllinux references the system arch on the PEP and does not explicitly take armv6 out.

auditwheel/policy/musllinux-policy.json
```diff
@@ -0,0 +1,48 @@
...
```
pvizeli, line 23:
```suggestion
      }
      "armv6l": {
      }
```

same

tests/integration/test_manylinux.py
```diff
@@ -232,526 +186,608 @@ def assert_show_output(manylinux_ctr, wheel, expected_tag, strict):
...
+        filenames = os.listdir(io_folder)
+        assert len(filenames) == 2
+        repaired_wheel = f'nonpy_rpath-0.1.0-{PYTHON_ABI}-{tag}.whl'
+        assert repaired_wheel in filenames
+        assert_show_output(manylinux_ctr, repaired_wheel, policy, False)
+
+        # Test the resulting wheel outside the manylinux container
+        docker_exec(docker_python, "pip install /io/" + repaired_wheel)
+        if 'musllinux' in tag:
+            pytest.xfail("libcrypt segfaults")
```
mayeut, line 609: This should also be investigated. I didn't have time to do much for now.

auditwheel/policy/musllinux-policy.json
mayeut: > Unlike manylinux, musllinux references the system arch on the PEP and does not explicitly take armv6 out.

[PEP 600 does not rule armv6 either](https://www.python.org/dev/peps/pep-0600/#core-definition) however, it requires  a new PEP for armv6 since it has multiple variants (at least soft-float & hard-float ABIs are both returned as `armv6l`). This was similar for armv7l but, given [PEP 599](https://www.python.org/dev/peps/pep-0599/), it represents the hard-float ABI when used for manylinux (& now musllinux)...

[PEP 656 specification](https://www.python.org/dev/peps/pep-0656/#specification) refers to the PEP 600 one where there's an example that I'll reproduce below:

> Example: The imaginary architecture LEG v7 has both big-endian and little-endian variants. Big-endian binaries require a big-endian system, and little-endian binaries require a little-endian system. But unfortunately, it's discovered that due to a bug in PEP 425, both variants use the same architecture tag, legv7. This makes it impossible to create a compliant manylinux_2_17_legv7 wheel: no matter what we do, it will crash on some user's systems. So, we write a new PEP defining architecture tags legv7le and legv7be; now we can ship manylinux LEG v7 wheels.

I wish a PEP had been the case for `armv7l` where `armhf` (as used by many distros) or `armv7hf` would have been clearer than `armv7l`.

tests/integration/test_manylinux.py
mayeut: The following code from musl seems relevant:
https://git.musl-libc.org/cgit/musl/tree/ldso/dynlink.c#n1078

auditwheel/policy/musllinux-policy.json
pvizeli: I think it's many what the Linux kernel will return, and that can be only armv7l and armv6l. Anyway, I'm fine with the answer; I will patch it locally for Home Assistant to build musl wheel for our usage.

mayeut: > I think it's what the Linux kernel will return, and that can be only armv7l and armv6l. 

You're right and that's because of this shortcoming of `distutils.util.get_platform()` that there are such patch in wheel & packaging:

https://github.com/pypa/wheel/blob/0b29b695ae458450217552978ad3094ed06a16d7/src/wheel/bdist_wheel.py#L47-L56

https://github.com/pypa/packaging/blob/e2bc8f6ef2d66f6eaaf522fcf70a0f203b734064/packaging/tags.py#L417-L427
https://github.com/pypa/packaging/blob/e2bc8f6ef2d66f6eaaf522fcf70a0f203b734064/packaging/_manylinux.py#L114-L119

tests/integration/test_manylinux.py
mayeut: It segfaults because it's not using libxcrypt but musl embedded `crypt` function.

The `crypt` & `crypt_r` function are loaded from `libc.musl-*.so.1` because musl does not honor symbol version (the symbol that should be linked according to readelf is `crypt@@XCRYPT_2.0`). The musl `crypt` function just segfaults when passing a `NULL` salt.
libxcrypt should be removed from the musllinux image.

tests/integration/test_manylinux.py
mayeut: fixed in `lddtree.py` even though it should be cleaned up a bit there.

mayeut:
@lkollar, I split the PR in 2:
- this one to add musslinux support
- #317 for integration tests once we have musllinux images

I think it's now good enough for review even though, as stated earlier, a refactor of the whole policy selection/loading shall be reworked later on.

auditwheel/lddtree.py
```diff
@@ -195,11 +198,31 @@ def load_ld_paths(root: str = '/', prefix: str = '') -> Dict[str, List[str]]:
...
-    ldpaths['conf'].extend(['/lib', '/lib64/', '/usr/lib', '/usr/lib64'])
+    libc = get_libc()
+    if libc == Libc.MUSL:
+        # from https://git.musl-libc.org/cgit/musl/tree/ldso
+        # /dynlink.c?id=3f701faace7addc75d16dea8a6cd769fa5b3f260#n1063
+        ld_musl = list(Path(root + prefix + '/etc').glob("ld-musl-*.path"))
+        assert len(ld_musl) <= 1
+        if len(ld_musl) == 0:
+            ldpaths['conf'] = [
+                root + '/lib',
```
lkollar, line 209: I wonder if the prefix should be taken into account here.

auditwheel/lddtree.py
```diff
@@ -195,11 +198,31 @@ def load_ld_paths(root: str = '/', prefix: str = '') -> Dict[str, List[str]]:
...
+        else:
+            ldpaths['conf'] = []
+            for ldpath in ld_musl[0].read_text().split(':'):
+                ldpath_stripped = ldpath.strip()
+                if ldpath_stripped == "":
+                    continue
+                ldpaths['conf'].append(root + ldpath_stripped)
+    else:
+        # Load up /etc/ld.so.conf.
+        ldpaths['conf'] = parse_ld_so_conf(root + prefix + '/etc/ld.so.conf',
```
lkollar, line 222:
```suggestion
        ldpaths['conf'] = parse_ld_so_conf(str(root_prefix / 'etc/ld.so.conf'),
```

auditwheel/lddtree.py
```diff
@@ -195,11 +198,31 @@ def load_ld_paths(root: str = '/', prefix: str = '') -> Dict[str, List[str]]:
...
-    # Load up /etc/ld.so.conf.
-    ldpaths['conf'] = parse_ld_so_conf(root + prefix + '/etc/ld.so.conf',
-                                       root=root)
-    # the trusted directories are not necessarily in ld.so.conf
-    ldpaths['conf'].extend(['/lib', '/lib64/', '/usr/lib', '/usr/lib64'])
+    libc = get_libc()
+    if libc == Libc.MUSL:
+        # from https://git.musl-libc.org/cgit/musl/tree/ldso
+        # /dynlink.c?id=3f701faace7addc75d16dea8a6cd769fa5b3f260#n1063
+        ld_musl = list(Path(root + prefix + '/etc').glob("ld-musl-*.path"))
```
lkollar, line 205:
```suggestion
        root_prefix = Path(root) / prefix
        ld_musl = list((root_prefix / 'etc').glob("ld-musl-*.path"))
```

auditwheel/lddtree.py
```diff
@@ -195,11 +198,31 @@ def load_ld_paths(root: str = '/', prefix: str = '') -> Dict[str, List[str]]:
...
+                root + '/usr/local/lib',
+                root + '/usr/lib'
+            ]
+        else:
+            ldpaths['conf'] = []
+            for ldpath in ld_musl[0].read_text().split(':'):
+                ldpath_stripped = ldpath.strip()
+                if ldpath_stripped == "":
+                    continue
+                ldpaths['conf'].append(root + ldpath_stripped)
```
lkollar, line 219: Prefix is not taken into account here either but I'm not actually sure how the prefix logic is supposed to work in this function so maybe it's fine.

lkollar: This could be:
```suggestion
                ldpaths['conf'].append(str(root_prefix / ldpath_stripped))
```

auditwheel/lddtree.py
mayeut: I'm not sure. I guess it won't matter until cross-repairing is a thing.
PS: for glibc, some folders are appended without either root or prefix.

auditwheel/lddtree.py
mayeut: I'd rather not touch the glibc code path for now. We can get this cleaned up later.

auditwheel/lddtree.py
mayeut: Same comment as earlier, it won't matter until cross-repairing is a thing.
It mimics `parse_ld_so_conf` where only `root` is taken into account.
The `prefix` seems to only be used for the configuration on glibc so I did the same on musl for now.

mayeut:
@lkollar,

are my answers enough for you to finish the review ?

lkollar (approved):

auditwheel/policy/musllinux-policy.json
andypost: Faced error building for this platform and used hack from https://github.com/PyO3/maturin/pull/2942#discussion_r2703710640

Any idea why it was removed?

PR #324, opened 2021-08-22
feature: replace `unzip` usage with Python's `zipfile`

### LINKED ISSUE #258: Replace `unzip` usage with Python's `zipfile`

mayeut:
This removes the need to have `unzip` installed.

fixes #258
closes #312


mayeut:
`tests/unit/test-permissions.zip.xz` created with:
```
mkdir test-permissions
pushd test-permissions 
for i in 0 1 2 3 4 5 6 7; do
	for j in 0 1 2 3 4 5 6 7; do
		for k in 0 1 2 3 4 5 6 7; do
			perm=${i}${j}${k}
			mkdir ${perm}.d
			touch ${perm}.d/f
			chmod ${perm} ${perm}.d
			touch ${perm}.f
			chmod ${perm} ${perm}.f
		done
	done
done
sudo zip -r -0 ../test-permissions.zip .
popd
sudo rm -rf test-permissions
xz -z -e -f test-permissions.zip
```

tests/unit/test_tools.py
```diff
@@ -43,3 +48,41 @@ def test_environment_action_invalid_env(monkeypatch):
             dest='PLAT',
             choices=choices,
             default='manylinux1')
+
+
+def _write_test_permissions_zip(path):
+    source_zip_xz = Path(__file__).parent / "test-permissions.zip.xz"
```
lkollar, line 54: Why is this file compressed twice?

lkollar (approved):
Looks good to me. A question on the tests though: why not create the test zip fie in the test suite itself instead of checking the file in? That would make it clearer how the file is created and easier to maintain in case we have to change the input in the future.

tests/unit/test_tools.py
mayeut: the zip is 215kB, with xz compression, goes down to 8kB

mayeut:
> A question on the tests though: why not create the test zip fie in the test suite itself instead of checking the file in? 

I wanted a single source for that file without wondering if there was an implementation detail in python x.y that would make things flaky.
I can still try to implement this in python & add some kind of check to make sure everything is as expected in the zip file.

auvipy (approved):

PR #329, opened 2021-08-28
chore: run black

mayeut:
Reformat the code base with black.
It provides a known look & feel compared to other pypa projects (e.g. build, cibuildwheel, packaging, pip, ...)

### DIFF
--- a/auditwheel/__main__.py
+++ b/auditwheel/__main__.py
@@ -2,5 +2,5 @@ import sys
 
 from .main import main
 
-if __name__ == '__main__':
+if __name__ == "__main__":
     sys.exit(main())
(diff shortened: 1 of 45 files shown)


lkollar (approved):
Can we also use a `.git-blame-ignore-revs` file to ignore the commit in `git blame`? See https://www.moxio.com/blog/43/ignoring-bulk-change-commits-with-git-blame.

lkollar (changes requested):
We should also add black to the lint rule in tox.

mayeut:
> We should also add black to the lint rule in tox.

This will be handled by #331

mayeut:
> Can we also use a .git-blame-ignore-revs file to ignore the commit in git blame?

Depending how it'll be merged, the commit might change. better do this after it's merged.

lkollar (approved):

PR #332, opened 2021-08-28
chore: use `src` layout

mayeut:
Allows better test isolation & thus, test the sdist / wheel without the legacy layout interfering.

depends on #331


lkollar (commented):
LGTM but I can't figure out why the coverage has dropped so much. Any ideas @mayeut?

mayeut:
>  I can't figure out why the coverage has dropped so much. Any ideas @mayeut?

Taking a closer look at this, it seems all manylinux tests are skipped. I need to take a closer look to see what's wrong here...

mayeut:
> > I can't figure out why the coverage has dropped so much. Any ideas @mayeut?

> Taking a closer look at this, it seems all manylinux tests are skipped. I need to take a closer look to see what's wrong here...

This now fixed and coverage is back to no diff which is expected.

lkollar (approved):

PR #334, opened 2021-08-29
feat: add `libz.so.1` to whitelisted libraries

### LINKED ISSUE #152: We should probably allow libz linkage, and maybe libexpat

### LINKED ISSUE #161: libz.so.1.2.3 on CentOS 6 causes repair to segfault

mayeut:
This also requires to blacklist some symbols from `libz.so.1`.
There was some breakage in ABI compatibility for this library.
It's seems unlikely that those symbols are used in the wild.

Fix #152
Fix #161


auditwheel/policy/manylinux-policy.json
```diff
@@ -29,8 +32,12 @@
          "libc.so.6", "libnsl.so.1", "libutil.so.1", "libpthread.so.0",
          "libX11.so.6", "libXext.so.6", "libXrender.so.1", "libICE.so.6",
          "libSM.so.6", "libGL.so.1", "libgobject-2.0.so.0",
-         "libgthread-2.0.so.0", "libglib-2.0.so.0", "libresolv.so.2"
-     ]},
+         "libgthread-2.0.so.0", "libglib-2.0.so.0", "libresolv.so.2",
+         "libz.so.1"
+     ],
+     "blacklist": {
+       "libz.so.1": ["_dist_code", "_length_code", "_tr_align", "_tr_flush_block", "_tr_init", "_tr_stored_block", "_tr_tally", "bi_windup", "crc32_vpmsum", "crc_fold_512to32", "crc_fold_copy", "crc_fold_init", "deflate_copyright", "deflate_medium", "fill_window", "flush_pending", "gzflags", "inflate_copyright", "inflate_fast", "inflate_table", "longest_match", "slide_hash_sse", "static_ltree", "uncompress2", "x86_check_features", "x86_cpu_has_pclmul", "x86_cpu_has_sse2", "x86_cpu_has_sse42", "z_errmsg", "zcalloc", "zcfree"]
```
mayeut, line 39: We are repeating ourselves way too much here. Should probably go in a specific file that's common for manylinux &  musllinux.

auditwheel/elfutils.py
```diff
@@ -137,3 +137,27 @@ def is_subdir(path: str, directory: str) -> bool:
...
+        return {}
+    undef_symbols = set("*")
+    with open(path, "rb") as f:
+        elf = ELFFile(f)
+        section = elf.get_section_by_name(".dynsym")
+        if section is not None:
+            # look for all undef symbols
+            for sym in section.iter_symbols():
+                if sym["st_shndx"] == "SHN_UNDEF":
+                    undef_symbols.add(sym.name)
```
lkollar, line 155: nit: This could be made into a utility function to find undefined symbols. That would make this function more readable as well.

auditwheel/elfutils.py
```diff
@@ -137,3 +137,27 @@ def is_subdir(path: str, directory: str) -> bool:
...
+
+    result = {}
+    for lib in symbols:
+        intersection = set(symbols[lib]) & undef_symbols
```
lkollar, line 159:
```suggestion
    for lib, sym_list in symbols:
        intersection = set(sym_list) & undef_symbols
```

auditwheel/wheel_abi.py
```diff
@@ -285,12 +293,19 @@ def analyze_wheel_abi(wheel_fn: str) -> WheelAbIInfo:
         sym_tag,
         ucs_tag,
         pyfpe_tag,
+        blacklist_tag,
     )
 
 
 def update(d, u):
```
lkollar, line 300: Unrelated to this PR, but we should refactor this function.

mayeut:
> Where is the list of blacklisted symbols coming from? This is something we should definitely document to make sure we can keep the list up to date.

Everything comes from https://github.com/mayeut/pep600_compliance/pull/88 (The same repo was used to define policies for manylinux_2_24 & manylinux_2_27).
I crossed referenced this information with https://abi-laboratory.pro/index.php?view=timeline&l=zlib as well.

auditwheel/elfutils.py
mayeut: will do.

mayeut: done

lkollar (approved):

mayeut:
@vgod-dbx,

 > According to the manylinux policies, libz.so.1 should not be whitelisted.

You're probably talking about the previous [PEP 513](https://peps.python.org/pep-0513), [PEP 571](https://peps.python.org/pep-0571) & [PEP 599](https://peps.python.org/pep-0599). Those 3 PEP have been superseded by  the much more generic [PEP 600](https://peps.python.org/pep-0600) which does not list allowed / disallowed libraries & which states:
> `manylinux_${GLIBCMAJOR}_${GLIBCMINOR}_${ARCH}`
This tag is a promise: the wheel’s creator promises that the wheel will work on any mainstream Linux distro that uses glibc version `${GLIBCMAJOR}.${GLIBCMINOR}` or later, and where the `${ARCH}` matches the return value from distutils.util.get_platform(). (For more detail about architecture tags, see [PEP 425](https://peps.python.org/pep-0425).)

`libz.so.1` is available on all mainstream Linux distros & not whitelisting it caused some issues (ranging from "not working" to "security issue" on unmaintained grafted library).
`libz.so.1` is also part of the [LSB Core specification](https://refspecs.linuxfoundation.org/LSB_5.0.0/LSB-Core-generic/LSB-Core-generic/libz.html) which more or less means it should also be available on all Linux distros (i.e. not just the mainstream ones).

CPython itself relies on zlib for multiple modules (although, if you're using CPython and linking zlib statically, it does not depend on `libz.so.1`).

Your company should probably update its minimal runtime package with `libz.so.1`.

PR #343, opened 2021-10-21
Allow `auditwheel repair` to process multiple wheels

### LINKED ISSUE #62: Support multiple wheel files in repair command
Currently ``auditwheel repair`` command supports only one ``WHEEL_FILE`` argument, I thought it would be nice if it will be possible to give multiple wheels to repair, what do you think? It will be useful in my use case when I build a set of wheels some of which should be repaired. And if you consider it to be useful addition, I'll submit a patch.

tipabu:
Closes #62


auvipy (changes requested):
can you please investigate the reasons of build failures?

auvipy (approved):

PR #345, opened 2021-11-14
chore: test with python 3.10

mayeut:


lkollar (approved):

lkollar:
I wonder if the test failure is because numpy 1.19.2 has no wheel for Python 3.10.

mayeut:
> I wonder if the test failure is because numpy 1.19.2 has no wheel for Python 3.10.

We're building from sources so no, that's not the reason. I'll update all tests this week for them to pass.
e.g. manylinux1 does not have python 3.10, numpy should be updated (not sure there's a version which supports 3.6->3.10, might need to configure the numpy version depending on the python version).

Issue #352, opened 2021-12-08
libc version failed to be detected on CentOS8 based docker image


hjmallon:
My CI pipeline continued after this successfully, but we shouldn't be throwing warnings if `ldd /bin/ls` fails, as on CentOS 8 it seems to be a bash script.

Example warning in auditwheel repair:

```
Failed to determine libc version
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/auditwheel/musllinux.py", line 20, in find_musl_libc
    ldd = subprocess.check_output(["ldd", "/bin/ls"], errors="strict")
  File "/usr/lib64/python3.6/subprocess.py", line 356, in check_output
    **kwargs).stdout
  File "/usr/lib64/python3.6/subprocess.py", line 438, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['ldd', '/bin/ls']' returned non-zero exit status 1.
```

```
sh-4.4# ldd /bin/ls
	not a dynamic executable
sh-4.4# echo $?
1
# cat /bin/ls
#!/usr/bin/coreutils --coreutils-prog-shebang=ls
```

PR #353, opened 2021-12-12
Simplify musl platform check 

lkollar:
Closes: #352

### DIFF
--- a/src/auditwheel/musllinux.py
+++ b/src/auditwheel/musllinux.py
@@ -17,21 +17,12 @@ class MuslVersion(NamedTuple):
 
 def find_musl_libc() -> pathlib.Path:
     try:
-        ldd = subprocess.check_output(["ldd", "/bin/ls"], errors="strict")
-    except (subprocess.CalledProcessError, FileNotFoundError):
+        (dl_path,) = list(pathlib.Path("/lib").glob("libc.musl-*.so.1"))
+    except ValueError:
         LOG.error("Failed to determine libc version", exc_info=True)
         raise InvalidLibc
 
-    match = re.search(
-        r"libc\.musl-(?P<platform>\w+)\.so.1 "  # TODO drop the platform
-        r"=> (?P<path>[/\-\w.]+)",
-        ldd,
-    )
-
-    if not match:
-        raise InvalidLibc
-
-    return pathlib.Path(match.group("path"))
+    return dl_path
 
 
 def get_musl_version(ld_path: pathlib.Path) -> MuslVersion:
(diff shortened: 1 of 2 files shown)


auvipy (approved):

auvipy:
may be the tests need some adjustment?

lkollar:
The test coverage of musllinux.py is 100% but since there is less code now in the project, the overall coverage is slightly lower.

src/auditwheel/musllinux.py
```diff
@@ -40,7 +31,7 @@ def get_musl_version(ld_path: pathlib.Path) -> MuslVersion:
             [ld_path], check=False, errors="strict", stderr=subprocess.PIPE
         ).stderr
     except FileNotFoundError:
-        LOG.error("Failed to determine musl version", exc_info=True)
+        LOG.debug("Failed to parse musl version")
```
mayeut, line 34: This one shall probably still log the error.

src/auditwheel/musllinux.py
```diff
@@ -17,21 +17,12 @@ class MuslVersion(NamedTuple):
 
 def find_musl_libc() -> pathlib.Path:
     try:
-        ldd = subprocess.check_output(["ldd", "/bin/ls"], errors="strict")
-    except (subprocess.CalledProcessError, FileNotFoundError):
-        LOG.error("Failed to determine libc version", exc_info=True)
+        (dl_path,) = list(pathlib.Path("/lib").glob("*ld-musl-*.so.1"))
```
mayeut, line 20: I might have missed a previous commit but why the leading `*` ?
If it's not needed, it's probably better to be stricter.
Shall the architecture part of the name be checked for consistency  ?
With what's in https://github.com/pypa/auditwheel/blob/353c24250d66951d5ac7e60b97471a6da76c123f/src/auditwheel/policy/__init__.py#L82-L91

mayeut: Shall be look for `ld` or  `libc` ? I'd rather look for libc as it allows to use the same look-up in policies & here.

lkollar: I've updated this to look for libc instead. On musl libc is a symlink to the dynamic loader, but this should be fine.

I'm not sure about the architecture check. I think to reuse the `soname_map` in both places, we'd have to move `find_musl_libc` into the policy module to avoid a circular import. It also requires refactoring `_fixup_musl_libc_soname`. I think this change should be done separately.

mayeut: let's do this separately then.

Issue #321, opened 2021-08-13
auditwheel 5.1.0: policy.json not installed when dist-info created first / when running from git tarball


hroncok:
When I run the tests over installed auditwheel 4.0.0 from source, It fails with:

```
==================================== ERRORS ====================================
___________________ ERROR collecting tests/unit/test_main.py ___________________
tests/unit/test_main.py:4: in <module>
    from auditwheel.main import main
../../BUILDROOT/auditwheel-4.0.0-1.fc35.x86_64/usr/lib/python3.10/site-packages/auditwheel/main.py:11: in <module>
    from . import main_repair
../../BUILDROOT/auditwheel-4.0.0-1.fc35.x86_64/usr/lib/python3.10/site-packages/auditwheel/main_repair.py:4: in <module>
    from .policy import (load_policies, get_policy_by_name, get_policy_name,
../../BUILDROOT/auditwheel-4.0.0-1.fc35.x86_64/usr/lib/python3.10/site-packages/auditwheel/policy/__init__.py:59: in <module>
    with open(join(dirname(abspath(__file__)), 'policy.json')) as f:
E   FileNotFoundError: [Errno 2] No such file or directory: '/builddir/build/BUILDROOT/auditwheel-4.0.0-1.fc35.x86_64/usr/lib/python3.10/site-packages/auditwheel/policy/policy.json'
__________________ ERROR collecting tests/unit/test_policy.py __________________
tests/unit/test_policy.py:5: in <module>
    from auditwheel.policy import get_arch_name, get_policy_name, \
../../BUILDROOT/auditwheel-4.0.0-1.fc35.x86_64/usr/lib/python3.10/site-packages/auditwheel/policy/__init__.py:59: in <module>
    with open(join(dirname(abspath(__file__)), 'policy.json')) as f:
E   FileNotFoundError: [Errno 2] No such file or directory: '/builddir/build/BUILDROOT/auditwheel-4.0.0-1.fc35.x86_64/usr/lib/python3.10/site-packages/auditwheel/policy/policy.json'
__________________ ERROR collecting tests/unit/test_repair.py __________________
tests/unit/test_repair.py:5: in <module>
    from auditwheel.repair import append_rpath_within_wheel
../../BUILDROOT/auditwheel-4.0.0-1.fc35.x86_64/usr/lib/python3.10/site-packages/auditwheel/repair.py:17: in <module>
    from .policy import get_replace_platforms
[...]

PR #362, opened 2022-01-03
fix: include tests in SDist

mayeut:
Required for downstream packagers.

Reported in #321 
Tests were removed from SDist in 5.1.0


hroncok:
Thanks.

hroncok:
It works as expected. Thanks.

PR #364, opened 2022-01-04
Allow setuptools_scm 6.0

hroncok:
We have setuptools_scm 6.0 in Fedora 35 and I've tried to build auditwheel with it -- it worked.

I am not entirely sure why 6.2 was used here, maybe just "to be safe"... ?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=45", "wheel", "setuptools_scm>=6.2"]
+requires = ["setuptools>=45", "wheel", "setuptools_scm>=6.0"]
 build-backend = "setuptools.build_meta"
 
 [tool.setuptools_scm]


auvipy (approved):

mayeut:
> I am not entirely sure why 6.2 was used here, maybe just "to be safe"... ?

Because that's what appears when you look at `setuptools_scm` README for `pyproject.toml` configuration: https://github.com/pypa/setuptools_scm

Looking at https://github.com/pypa/setuptools_scm/tree/v6.1.1 mentions `setuptools_scm[toml]>=6.0` rather than `setuptools_scm>=6.0`.
The `tomli` dependency is now always required starting with the 6.2 version and they thus changed the README at that point.

Given that you already have a patch that's working on your setup, that fc36 doesn't need this patch, I'm not really in favor of lowering the version and introducing the extra (this probably just asking for things to break in the future) unless you have other arguments to present.

hroncok:
That's reasonable. I guess the fc35 build only works better cause we pull in toml by other means, I'll make sure to fix that in Fedora.

PR #368, opened 2022-02-05
Add --exclude option to auditwheel repair

### LINKED ISSUE #76: Whitelisting of libraries shipped by other wheels
In `turbodbc` we link against `pyarrow` in such a fashion that `turbodbc` picks up the Arrow libraries via a relative RPATH in the same virtualenv. Thus we don't need to include them in the `manylinux1` wheel of Turbodbc. With auditwheel's current behaviour they are sadly grafted into turbodbc's wheel. To fix this, two options come to my mind:

 * whilelist all libraries that are contained in the virtualenv
 * explicitly whitelist some libraries via a commandline option

I'm happy to implement one of these approaches but I would like to hear what the preferred one would be.

### LINKED ISSUE #241: Do not place OpenGL libraries into wheels on manylinux*

### LINKED ISSUE #391: Allow not copying `.so` files provided by other PyPI packages (e.g. `libtorch*.so` from `torch`)
`auditwheels` copies all external `.so` files into `<packagename>.lib` folder. Then packs them into the wheel file. Sometimes the current package (the developer what to audit) only has very small C++ extension, but it references a large `.so` file. For example, the user builds a C++ extension for PyTorch, it references `libtorch*.so` (some files up to **0.5GiB+**). This makes the final wheel file extremely large (~2GiB).

We can add an option that if the `.so` files are located inside `site-packages`, then do not copy them but update the `rpath` of the target.

For example, the user builds a C++ extension for PyTorch, it references `libtorch_python.so`. `auditwheel` can update the `rpath` to:

```text
$ORIGIN/../torch/lib  # the relative path should be calculated carefully
```

where `$ORIGIN/..` is `site-packages`. And then adds `torch` to the install dependency.

martinRenou:
This is a follow-up of #310 by @rossant, only keeping the `exclude` option as suggested by @mayeut

Closes #76
Closes #241
Closes #310
Fixes #391

Original PR description:

----

This is a quick fix to solve an issue I have with my project and for which auditwheel seemed to help. I don't know if this approach is suitable to other users of auditwheel.

I develop a C graphics library ([datoviz](https://github.com/datoviz/datoviz/)) that depends on Vulkan and comes with Cython bindings. https://github.com/datoviz/datoviz/issues/13 for Linux (at least Ubuntu 20.04, that would be a start).

1. As a first step, I compile the C project and get a libdatoviz.so shared library.
2. Then, I build the Cython library, which has a dynamic dependency to libdatoviz.
3. I'd like to bundle both the Cython extension module, and libdatoviz, in the same wheel.
4. I tried to use auditwheel repair. It included dozens of other dependencies in the wheel, including libvulkan and many graphics-related dependent libraries. The compiled wheel installs properly, but it doesn't work properly (there are issues related to windowing, GPU access, etc). Related: https://github.com/pypa/auditwheel/issues/241[](https://github.com/rossant)
5. I made some changes to auditwheel (this pull request) to exclude libvulkan and a few other libraries. Some issues were fixed, but not others.
6. As another approach, I tried to bundle just libdatoviz, and no other library in the wheel. That seems to work, at least on my computer. Chances are that the wheel may not work on other Linux-based operating systems though.
Any help would be appreciated, regarding either this pull request, and/or a way to solve the issue I'm facing. Thanks!

### DIFF
--- a/src/auditwheel/main_repair.py
+++ b/src/auditwheel/main_repair.py
@@ -89,6 +89,14 @@ wheel will abort processing of subsequent wheels.
         default=False,
     )
     p.add_argument(
+        "--exclude",
+        dest="EXCLUDE",
+        help="Exclude SONAME from grafting into the resulting wheel "
+        "(can be specified multiple times)",
+        action="append",
+        default=[],
+    )
+    p.add_argument(
         "--only-plat",
         dest="ONLY_PLAT",
         action="store_true",
@@ -169,6 +177,7 @@ def execute(args, p):
             out_dir=args.WHEEL_DIR,
             update_tags=args.UPDATE_TAGS,
             patcher=patcher,
+            exclude=args.EXCLUDE,
             strip=args.STRIP,
         )
 
--- a/src/auditwheel/repair.py
+++ b/src/auditwheel/repair.py
@@ -37,6 +37,7 @@ def repair_wheel(
     out_dir: str,
     update_tags: bool,
     patcher: ElfPatcher,
+    exclude: List[str],
     strip: bool = False,
 ) -> Optional[str]:
 
@@ -70,6 +71,11 @@ def repair_wheel(
             ext_libs = v[abis[0]]["libs"]  # type: Dict[str, str]
             replacements = []  # type: List[Tuple[str, str]]
             for soname, src_path in ext_libs.items():
+
+                if soname in exclude:
+                    logger.info(f"Excluding {soname}")
+                    continue
+
                 if src_path is None:
                     raise ValueError(
                         (
--- a/tests/integration/test_bundled_wheels.py
+++ b/tests/integration/test_bundled_wheels.py
@@ -67,6 +67,7 @@ def test_wheel_source_date_epoch(tmp_path, monkeypatch):
         UPDATE_TAGS=True,
         WHEEL_DIR=str(wheel_output_path),
         WHEEL_FILE=[str(wheel_path)],
+        EXCLUDE=[],
         cmd="repair",
         func=Mock(),
         prog="auditwheel",
--- a/tests/integration/test_manylinux.py
+++ b/tests/integration/test_manylinux.py
@@ -315,61 +315,49 @@ class Anylinux:
     @pytest.mark.skipif(
         PLATFORM != "x86_64", reason="Only needs checking on one platform"
     )
-    def test_repair_exclude(self, io_folder):
+    def test_repair_exclude(self, any_manylinux_container, io_folder):
         """Test the --exclude argument to avoid grafting certain libraries."""
 
-        policy = "manylinux_2_12"
-        policy_plat = f"{policy}_{PLATFORM}"
-        base = MANYLINUX_IMAGES[policy]
-        env = {"PATH": PATH[policy]}
+        policy, tag, manylinux_ctr = any_manylinux_container
 
-        with tmp_docker_image(
-            base,
-            [
-                'git config --global --add safe.directory "/auditwheel_src"',
-                "pip install -U pip setuptools pytest-cov",
-                "pip install -U -e /auditwheel_src",
-            ],
-            env,
-        ) as tmp_img:
-            with docker_container_ctx(tmp_img, io_folder, env) as container:
-                platform_tag = ".".join(
-                    [
-                        f"{p}_{PLATFORM}"
-                        for p in [policy] + POLICY_ALIASES.get(policy, [])
-                    ]
-                )
+        orig_wheel = build_numpy(manylinux_ctr, policy, io_folder)
+        assert orig_wheel == ORIGINAL_NUMPY_WHEEL
+        assert "manylinux" not in orig_wheel
 
-                orig_wheel = build_numpy(container, policy_plat, io_folder)
-                assert orig_wheel == ORIGINAL_NUMPY_WHEEL
-                assert "manylinux" not in orig_wheel
+        # Exclude libgfortran from grafting into the wheel
+        excludes = {
+            "manylinux_2_5_x86_64": ["libgfortran.so.1", "libgfortran.so.3"],
+            "manylinux_2_12_x86_64": ["libgfortran.so.3", "libgfortran.so.5"],
+            "manylinux_2_17_x86_64": ["libgfortran.so.3", "libgfortran.so.5"],
+            "manylinux_2_24_x86_64": ["libgfortran.so.3"],
+            "manylinux_2_28_x86_64": ["libgfortran.so.5"],
+            "musllinux_1_1_x86_64": ["libgfortran.so.5"],
+        }[policy]
 
-                # Exclude libgfortran from grafting into the wheel
-                output = docker_exec(
-                    container,
-                    [
-                        "auditwheel",
-                        "repair",
-                        "--plat",
-                        policy_plat,
-                        "--exclude",
-                        "libgfortran.so.5",
-                        "--only-plat",
-                        "-w",
-                        "/io",
-                        f"/io/{orig_wheel}",
-                    ],
-                )
+        repair_command = [
+            "auditwheel",
+            "repair",
+            "--plat",
+            policy,
+            "--only-plat",
+            "-w",
+            "/io",
+        ]
+        for exclude in excludes:
+            repair_command.extend(["--exclude", exclude])
+        repair_command.append(f"/io/{orig_wheel}")
+        output = docker_exec(manylinux_ctr, repair_command)
 
-        assert "Excluding libgfortran" in output
+        for exclude in excludes:
+            assert f"Excluding {exclude}" in output
         filenames = os.listdir(io_folder)
         assert len(filenames) == 2
-        repaired_wheel = f"numpy-{NUMPY_VERSION}-{PYTHON_ABI}-{platform_tag}.whl"
+        repaired_wheel = f"numpy-{NUMPY_VERSION}-{PYTHON_ABI}-{tag}.whl"
         assert repaired_wheel in filenames
 
-        # Make sure we don't have libatlas in the result
+        # Make sure we don't have libgfortran in the result
         contents = zipfile.ZipFile(os.path.join(io_folder, repaired_wheel)).namelist()
-        assert not any(x for x in contents if x.startswith("libatlas"))
+        assert not any(x for x in contents if "/libgfortran" in x)
 
     def test_build_wheel_with_binary_executable(
         self, any_manylinux_container, docker_python, io_folder
--- a/tests/integration/test_manylinux.py
+++ b/tests/integration/test_manylinux.py
@@ -312,6 +312,65 @@ class Anylinux:
         # at once in the same Python program:
         docker_exec(docker_python, ["python", "-c", "'import numpy; import foo'"])
 
+    @pytest.mark.skipif(
+        PLATFORM != "x86_64", reason="Only needs checking on one platform"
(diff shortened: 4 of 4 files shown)


src/auditwheel/main_repair.py
```diff
@@ -85,6 +85,12 @@ def configure_parser(sub_parsers):
         help="Strip symbols in the resulting wheel",
         default=False,
     )
+    p.add_argument(
+        "--exlude",
+        dest="EXCLUDE",
+        help="List of excluded so files as a comma-separated string",
```
mayeut, line 91: Not sure about the plural of `soname` or using this directly as a noun but it disambiguates wether we're talking file names/file paths or soname.
```suggestion
        help="List of excluded sonames as a comma-separated string",
```

src/auditwheel/repair.py
```diff
@@ -30,6 +30,18 @@
 ).match
 
 
+def _is_in_list(soname: str, items: Optional[List[str]]) -> Optional[str]:
+    for item in (items or []):
+        if item in soname:
```
mayeut, line 35: I think we should have an exact match. Why shouldn't we ?

src/auditwheel/repair.py
```diff
@@ -30,6 +30,18 @@
...
+            return item
+
+    return None
+
+
+def _filter(items: Optional[List[str]]) -> List[str]:
```
mayeut, line 41: It's a one liner used exactly once, having a function does not help readability and/or factorization here.

mayeut (changes requested):
Thanks for your contribution,
I left a couple remarks inlined.
The PR is failing CI:
- we do not have pre-commit.ci setup yet so please make sure to run `nox -s lint` in order to check/repair for linter issues. if you don't know it, [nox](https://nox.thea.codes/en/stable/) is a command-line tool that automates testing in multiple Python environments.
- codecov is complaining of a drop in coverage because there's no test associated with this new feature. Can you please add some tests for this ?

src/auditwheel/main_repair.py
```diff
@@ -85,6 +85,12 @@ def configure_parser(sub_parsers):
         help="Strip symbols in the resulting wheel",
         default=False,
     )
+    p.add_argument(
+        "--exlude",
```
alexsavulescu, line 89:
```suggestion
        "--exclude",
```

pramodk: this is great work @martinRenou ! we need this in hpc environment to exclude libraries like OpenMP/OpenACC.

As Alex suggested, naming option `exclude` is more obvious here? (or, this was just a typo)

martinRenou: This was a typo indeed, thanks for reviewing!

martinRenou:
Thanks for reviewing! I'll try to find the time to add tests and fix linting

src/auditwheel/repair.py
```diff
@@ -38,6 +50,7 @@ def repair_wheel(
     update_tags: bool,
     patcher: ElfPatcher,
     strip: bool = False,
+    exclude: Optional[List[str]] = None,
```
ferdonline, line 53: I would default this to the empty tuple so that we can simplify logic everywhere it's used
```suggestion
    exclude: List[str] = (),
```

src/auditwheel/repair.py
ferdonline: This could well be a lambda internal to `repair_wheel`:
```
_exclude_rule = lambda item: next((item in soname for item in exclude), None)
```
But I also wonder about partial string matching. We probably want very good control about whike libraries are excluded.
In case you're thinking about matching groups, then make it explicit, like mylibgroup* (check the glob module)

src/auditwheel/repair.py
martinRenou: Thanks!

leofang:
I think this is a nice addition. Any chance we can get this in sooner? 🙂

auvipy (dismissed):
please increase the test coverage

martinRenou:
I don't mind at all. Thanks a lot!

lkollar:
I haven't merged this yet, as I'm still unsure if it is a good idea to exclude arbitrary libraries. On one hand, it solves a real issue, and I don't see how else we could achieve this, but it also ends up violating the manylinux spec. There is a risk that this will cause confusion and bad user experience, and it makes it more likely that users will run into ABI incompatibilities due to version mismatches between the extension and its excluded dependencies. Another concern is that if PyPI wanted to block uploads of non-conforming wheels, this feature would make it more difficult to implement it (see https://github.com/pypi/warehouse/issues/5420).

I wonder if this needs a larger discussion with the PyPA community, since the scope is bigger here than just auditwheel. If we want to properly support wheels which require some special external libraries, that could require changes in tooling in other areas as well. I don't have the bandwidth to take this on, but if someone wanted to raise this on the PyPA Discourse, that would be a good start.

Any thoughts @mayeut?

mattip:
I think this would allow abusing the purpose of auditwheel. I commented on the original issue about not pulling `libOpenGL` shared objects into the wheels.

jakirkham:
> I wonder if this needs a larger discussion with the PyPA community, since the scope is bigger here than just auditwheel. If we want to properly support wheels which require some special external libraries, that could require changes in tooling in other areas as well. I don't have the bandwidth to take this on, but if someone wanted to raise this on the [Packaging Discourse](https://discuss.python.org/c/packaging/14), that would be a good start.

There was [PEP 668]( https://peps.python.org/pep-0668/ ), which AIUI was also trying to help this use case of having some libraries externally managed. This is the result of a few discussions in different threads over the years (including past PyCons):

* https://discuss.python.org/t/drawing-a-line-to-the-scope-of-python-packaging/883
* https://discuss.python.org/t/playing-nice-with-external-package-managers/1968
* https://discuss.python.org/t/pep-668-marking-python-base-environments-as-externally-managed/10302

mayeut:
> I commented on the original issue about not pulling libOpenGL shared objects into the wheels.

This would require a new ABI compatibility analysis to be done to ensure whitelisting is allowed per pep-600

> I think this would allow abusing the purpose of auditwheel.

Yes, it would allow abusing the purpose but allowing package maintainers to use this (responsibly) is a way not to have auditwheel maintainers on the critical path of many projects. As can be seen by the activity in the repo, I do not have that much time to spend on this project, I'm not sure about @lkollar, but I guess that's about the same.

see also my other comment in https://github.com/pypa/auditwheel/pull/310#issuecomment-849773358 

There are already many packages on PyPI abusing PEP600, this is no reason to do so but, IMHO, having an escape hatch is always a good thing to have. It's probably better to have this escape hatch and have the rest of auditwheel run its process than to have a bunch of hand-tagged manylinux1 wheels on PyPI.

One example of abuse (or let's say at least gray area) are GPU dependent packages. Those raised [some discussion](https://discuss.python.org/t/what-to-do-about-gpus-and-the-built-distributions-that-support-them/7125/58) but I'm not sure there's a real conclusion yet (& the discussion starter was package size rather than pep600 compliance).
For example, one of torch wheels is torch-1.12.1-cp310-cp310-manylinux1_x86_64.whl. At least one of the shared object in that wheel links against `libcuda.so.1` which is part of the GPU driver and should not be grafted (nor whitelisted per strict PEP600) thus can't be repaired by auditwheel today (& "this led" to an invalid manylinux1 tag - might be another reason entirely but sure would have been a blocker).

> There is a risk that this will cause confusion and bad user experience, and it makes it more likely that users will run into ABI incompatibilities due to version mismatches between the extension and its excluded dependencies. 

Agreed but my comment above applies: I think the pros outweigh the cons.


> Another concern is that if PyPI wanted to block uploads of non-conforming wheels, this feature would make it more difficult to implement it (see https://github.com/pypi/warehouse/issues/5420).

I think that if PyPI is to implement that feature, they should not block on missing shared objects. They can't know for sure they're not compliant unless doing a manual review. We can't even know for sure that a manylinux1 tagged wheel that embeds some manylinux2014+ only symbol versions is not in fact manylinux1 compliant although in 99.9% of cases, we'd be right to assume it should be tagged manylinux2014.

> I wonder if this needs a larger discussion with the PyPA community, since the scope is bigger here than just auditwheel. If we want to properly support wheels which require some special external libraries, that could require changes in tooling in other areas as well. I don't have the bandwidth to take this on, but if someone wanted to raise this on the [Packaging Discourse](https://discuss.python.org/c/packaging/14), that would be a good start.

While I personally think this should be merged, some discussions should be raised but I don't have the bandwidth to take this on either.

leofang:
> One example of abuse (or let's say at least gray area) are GPU dependent packages. Those raised [some discussion](https://discuss.python.org/t/what-to-do-about-gpus-and-the-built-distributions-that-support-them/7125/58) but I'm not sure there's a real conclusion yet (& the discussion starter was package size rather than pep600 compliance).

Speaking from my personal experience, the whitelisting discussion here is intimately related to the package size discussion. They are the different facets of the same problem.

For GPU packages (or any packages that rely on vender-provided, proprietary prebuilt binaries) without whitelisting they only have the following options:
1. dlopen
2. static linking or auditwheel repair

Option 1 is technically still in violation of the auditwheel spirit, because it's still dynamic linking and could allow a non-self-contained wheel. It also requires a careful runtime check to ensure the correct shared library is dlopen'd. 

Option 2, on the other hand, leads to significantly increased package sizes and, in particular for proprietary packages, potential violation of the vendor's EULA (some vendors put very strict redistribution limitation). Furthermore, to adopt either option requires the downstream package providers to invest significant effort.

I am +1 for whitelisting. The only (minor) disagreement I have with @mayeut's above comment is I see no cons, only pros 🙂

mayeut:
Thanks for your feedback @rgommers.

I edited my comment about a community-wide discussion. I think you're absolutely right on that one. I guess @mattip changed [his view about not wanting this PR to be merged](https://github.com/pypa/auditwheel/pull/368#issuecomment-1217510548).
I will let @lkollar some time to answer your question though.

mattip:
Yes, I changed my view as I learned more about the pain points mentioned above: practicality sometimes beats purity.

mayeut (approved):
I'll let another week to @lkollar to comment. I'll merge this next week otherwise.

XuehaiPan:
I think this PR also requires an option to update the `rpath` as:

- #392

did. I.e., add the relative path of the skipped libraries in `rpath`. Otherwise, the wheel may be broken for missing libraries.

rgommers:
> I.e., add the relative path of the skipped libraries in `rpath`. Otherwise, the wheel may be broken for missing libraries.

No, please do not do that, it is not going to work. See https://github.com/pypa/auditwheel/issues/391#issuecomment-1272349511 for why, and further down in that discussion for a more robust solution: https://github.com/pypa/auditwheel/issues/391#issuecomment-1272380159.

leofang:
The `rpath` hack does not work for system dependencies (OpenGL, CUDA, Vulkan, ...), which is a wide realm that this PR aims to cover, so please do not do that.

mayeut:
Let's get this merged.
Thanks @martinRenou for the PR.
Thanks to all for your inputs.

lkollar:
Sorry for not answering earlier @mayeut. I was going to agree with merging this though. Thanks!

PR #369, opened 2022-02-05
test: add manylinux_2_28 tests

mayeut:
draft just to see if we're seeing complaints with `GLIBCXX_3.4.25` symbol version.


mayeut:
webhooks & GHA currently out of order: https://www.githubstatus.com/incidents/vvthhf8gxt80
let's wait to see how it goes...

tests/integration/test_manylinux.py
```diff
@@ -708,9 +708,7 @@ def test_glibcxx_3_4_25(self, any_manylinux_container, docker_python, io_folder)
         assert orig_wheel.startswith("testentropy-0.0.1")
 
         # Repair the wheel using the appropriate manylinux container
-        repair_command = (
-            f"auditwheel repair --plat {policy} --strip -w /io /io/{orig_wheel}"
```
mayeut: we should probably not try to strip system libraries that get grafted.
the tests fails on musllinux because libstdc++ and/or libgcc_s get stripped.

PR #372, opened 2022-02-17
Corrected link syntax

radarhere:
At https://github.com/pypa/auditwheel/blob/main/CHANGELOG.md#320, the links within "User-facing changes" are incorrectly formatted.

For example, the first one should be a link to `https://github.com/pypa/auditwheel/pull/237`, displaying simply as `#237`, but instead it's the opposite - a link to `#237` displaying as `https://github.com/pypa/auditwheel/pull/237`


mayeut:
Thanks

PR #374, opened 2022-02-23
feature: allow specifying additional library paths (fixes: #373)

### LINKED ISSUE #373: Handling additional library directories from cibuidwheel
In several projects I maintain (e.g. `PyAV`, `aiortc`, `aioquic`..) I need to build additional libraries which then get shipped with the wheels. These additional libraries are built with a custom prefix (`/tmp/vendor`) so as to keep them explicitly separate from the system libraries.

In order to get `auditwheel` to pick up the additional libraries, I am forced to add `/tmp/vendor/lib` to `LD_LIBRARY_PATH`. However this is very error prone, as it can break other parts of the build.

Would you be open to a PR which introduces an `--add-path` command-line flag in order to pass additional directories to `auditwheel` during it's search for libraries? The name of the option was chosen so as to match the one used by `delvewheel` [1], the Windows counterpart of `auditwheel`.

[1]  https://github.com/adang1345/delvewheel

jlaine:
Add new --add-path option to auditwheel repair to search for libraries
in additional directories. This avoid having to manipulate
LD_LIBRARY_PATH in the calling process (e.g. cibuildwheel), as this can
have undesirable side-effects.

The name of the option is chosen to match the one provided by
`delvewheel` on Windows.

fixes #373

### DIFF
--- a/src/auditwheel/main_repair.py
+++ b/src/auditwheel/main_repair.py
@@ -1,5 +1,6 @@
 import argparse
 import logging
+import os
 from os.path import abspath, basename, exists, isfile
 
 from auditwheel.patcher import Patchelf
@@ -92,6 +93,12 @@ below.
         help="Do not check for higher policy compatibility",
         default=False,
     )
+    p.add_argument(
+        "--add-path",
+        dest="PATHS",
+        default="",
+        help=f"Additional path(s) to search for libraries, {os.pathsep!r}-delimited",
+    )
     p.set_defaults(func=execute)
 
 
@@ -105,6 +112,12 @@ def execute(args, p):
         p.error("cannot access %s. No such file" % args.WHEEL_FILE)
 
     logger.info("Repairing %s", basename(args.WHEEL_FILE))
+
+    if args.PATHS:
+        library_path = args.PATHS
+        if "LD_LIBRARY_PATH" in os.environ:
+            library_path += os.pathsep + os.environ["LD_LIBRARY_PATH"]
+        os.environ["LD_LIBRARY_PATH"] = library_path
 
     if not exists(args.WHEEL_DIR):
         os.makedirs(args.WHEEL_DIR)
--- a/tests/integration/test_bundled_wheels.py
+++ b/tests/integration/test_bundled_wheels.py
@@ -62,6 +62,7 @@ def test_wheel_source_date_epoch(tmp_path, monkeypatch):
     args = Namespace(
         LIB_SDIR=".libs",
         ONLY_PLAT=False,
+        PATHS="",
         PLAT="manylinux_2_5_x86_64",
         STRIP=False,
         UPDATE_TAGS=True,


jlaine:
NOTE: this option would actually also make sense for the `lddtree` and `show` commands but I was unsure how to add it to multiple commands without duplicating code.

src/auditwheel/main_repair.py
```diff
@@ -92,6 +93,12 @@ def configure_parser(sub_parsers):
         help="Do not check for higher policy compatibility",
         default=False,
     )
+    p.add_argument(
+        "--add-path",
+        dest="PATHS",
+        default="",
+        help=f"Additional path(s) to search for libraries, {os.pathsep!r}-delimited",
+    )
```
mayeut, line 101: I understand that you wanted to use the same name as `delvewheel` which seems like a good thing however, it feels more like `-L`/`--library-path` or `-rpath-link` to mimic a bit more what feels like `ld` would be doing. I'd probably go with `--library-path`, allowed to be specified multiple times. Unfortunately, `-L` is already used for something else...

I'd be in favor of having the option specified multiple times with a single path each time rather than a single string option.

@lkollar, any thoughts on the name ? (Once it gets in, we won't be able to change it without breaking users).

mayeut (changes requested):
Thanks for your contribution.

> NOTE: this option would actually also make sense for the lddtree and show commands but I was unsure how to add it to multiple commands without duplicating code.

This is definitely needed. Probably requires a small helper file to setup the correct parser argument to all three & fix up LD_LIBRARY_PATH in a single place.

> Note that all of the tests in tests/integration/test_manylinux.py which currently manipulate LD_LIBRARY_PATH could be rewritten to use this new option. Would you like me to do this?

For part of them yes, It'd be good to keep some of them using the environment variable directly in order to ensure this won't get broken in the future.

src/auditwheel/main_repair.py
jlaine: @lkollar you're not missing anything, this is accurate, it can indeed be achieved by manipulating LD_LIBRARY_PATH. I hadn't realized that the "repair wheel command" was invoked in a shell, making it possible to alter environment variables. Now I'm not sure whether we actually want this option..

abitrolly: @jlaine I guess too many people do not realize that LD_LIBRARY_PATH is a solution for `ValueError: Cannot repair wheel, because required library "xxx.so" could not be located`, and that's exactly the reason why this PR is needed.

> Unfortunately, `-L` is already used for something else...

@mayeut and I can't say I didn't try it, because there was no other option.

`--add-path` is good for parity with `delvewheel` https://github.com/adang1345/delvewheel#additional-options

I would also add the pointer to it the error message, so that instead of plain `ValueError` there is a hint what to do next.

abitrolly: Actually the proper fix is to propagate the option values to `lddtree` here. It already parses LD_LIBRARY_PATH, so no need to handle that.

https://github.com/pypa/auditwheel/blob/55bc5ee7de795819fccfd6468129c1041bcb9ac7/src/auditwheel/wheel_abi.py#L83

zoj613:
What the progress on this? I have ran into this issue repairing wheels that were built linking to libraries installed bia conda.

mwtoews:
Would this PR resolve [limitation #1](https://github.com/pypa/auditwheel#limitations)? For instance, I'd like to create a package using pure `ctypes`.

mhsmith:
#597 is an alternate approach using an AUDITWHEEL_LD_LIBRARY_PATH environment variable.

PR #382, opened 2022-06-21
Add -vv TRACE logging level

abitrolly:
To reduce the output of policy checker unless requested.

### DIFF
--- a/src/auditwheel/main.py
+++ b/src/auditwheel/main.py
@@ -47,10 +47,13 @@ def main() -> Optional[int]:
     args = p.parse_args()
 
     logging.disable(logging.NOTSET)
-    if args.verbose >= 1:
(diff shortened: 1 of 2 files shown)


abitrolly:
`TRACE` level can probably be defined in the root `__init__.py` for convenience https://stackoverflow.com/questions/2183233/how-to-add-a-custom-loglevel-to-pythons-logging-facility/35804945#35804945

abitrolly:
The output from `versioned_symbols_policy` is overly verbose, because it is called several times and each time iterates over all JSON policies, from here.

https://github.com/pypa/auditwheel/blob/55bc5ee7de795819fccfd6468129c1041bcb9ac7/src/auditwheel/wheel_abi.py#L221-L223

In my case it is 64 times just for policy check without TRACE output.

Issue #386, opened 2022-07-22
Add runtime dependency on patchelf-pypi


lkollar:
The [patchelf-pypi](https://github.com/mayeut/patchelf-pypi) project ships up to date `patchelf` binaries. Adding this to `install_requires` would mean that users don't have to track down and install the package with a different package manager.

Any thoughts @mayeut? Any reason not to make this a dependency?

mayeut:
> Any reason not to make this a dependency?

There are a few reasons:
- If I'm not mistaken, `pipx` installed `auditwheel` wouldn't work out-of-the-box without care (need to expose the entry point of a dependency which is not the default behavior). This could be solved by reworking the `Patchelf` class.
- `patchelf` is only required for the `repair` command => this means a `auditwheel show` or a future `auditwheel check` could be run on any platform (not only linux for a check command) without patchelf which will probably never have macOS or Windows wheels.
- OS packagers will want to remove this dependency and use their own version of patchelf (like they are doing today e.g. in Fedora)
- Someone using a custom version of patchelf will need to go the extra mile to use it

Maybe a good first middle ground would be to add the dependency as an extra and update the documentation accordingly ?
e.g. `pip install auditwheel[patchelf]` or even just `pip install auditwheel patchelf` with no extra.

lkollar:
> * `patchelf` is only required for the `repair` command => this means a `auditwheel show` or a future `auditwheel check` could be run on any platform (not only linux for a check command) without patchelf which will probably never have macOS or Windows wheels.

This could be solved by making `patchelf` a platform-specific dependency when/if we add support for other platforms.



> * OS packagers will want to remove this dependency and use their own version of patchelf (like they are doing today e.g. in Fedora)

Good point. I agree that this is problematic.



> Maybe a good first middle ground would be to add the dependency as an extra and update the documentation accordingly ?
> e.g. `pip install auditwheel[patchelf]` or even just `pip install auditwheel patchelf` with no extra.

Good idea. And this is pretty simple to implement. We could also note this in the error message if auditwheel cannot find patchelf.

bigcat88:
patchelf-pypi do not provide wheels for Alpine at all and for debian\alpine with armv7 CPU.
will adding those dependency will try to install patchelf-pypi from source even if patchelf was installed with package manager or builded from source?

mayeut:
> patchelf-pypi do not provide wheels for Alpine

yes it does

> patchelf-pypi do not provide for debian\alpine with armv7 CPU

[...]

PR #387, opened 2022-07-22
Add runtime dependency on patchelf

lkollar:
Use the PyPI version so users don't have to install this with the system
package manager.

Closes: #386


mayeut:
closing per comments in #386

PR #392, opened 2022-08-06
feat: do not copy libraries in site-packages

### LINKED ISSUE #391: Allow not copying `.so` files provided by other PyPI packages (e.g. `libtorch*.so` from `torch`)

XuehaiPan:
Add a new option `--no-copy-site-libs` to skip libraries in site-packages for `auditwheel repair`.

This feature will significantly reduce the wheel file size if the referenced libraries are shipped with other PyPI packages.

**NOTE:** The developers may have their own risks in using this feature (e.g. they need to add the skipped library as an install dependency in `setup.py`).

Resolves #391

### DIFF
--- a/src/auditwheel/main_repair.py
+++ b/src/auditwheel/main_repair.py
@@ -105,8 +105,8 @@ wheel will abort processing of subsequent wheels.
         help=(
             "Do not copy libraries located in the site-packages directory from "
             "other packages. Just update the `rpath` only. The developer will "
-            "need to ensure that the skipped libraries are listed the package "
-            "install dependencies."
+            "need to ensure that the skipped libraries are listed in the "
+            "package install dependencies."
         ),
         default=True,  # default: COPY_SITE_LIBS=True
     )
--- a/src/auditwheel/main_repair.py
+++ b/src/auditwheel/main_repair.py
@@ -49,8 +49,10 @@ wheel will abort processing of subsequent wheels.
         metavar="PLATFORM",
         env="AUDITWHEEL_PLAT",
         dest="PLAT",
-        help="Desired target platform. See the available platforms under the "
-        f'PLATFORMS section below. (default: "{highest_policy}")',
+        help=(
+            "Desired target platform. See the available platforms under the "
+            "PLATFORMS section below. (default: %(default)r)"
+        ),
         choices=policy_names,
         default=highest_policy,
     )
@@ -59,7 +61,8 @@ wheel will abort processing of subsequent wheels.
         "--lib-sdir",
         dest="LIB_SDIR",
         help=(
-            "Subdirectory in packages to store copied libraries." ' (default: ".libs")'
+            "Subdirectory in packages to store copied libraries. "
+            "(default: %(default)r)"
         ),
         default=".libs",
     )
@@ -68,7 +71,7 @@ wheel will abort processing of subsequent wheels.
         "--wheel-dir",
         dest="WHEEL_DIR",
         type=abspath,
-        help=("Directory to store delocated wheels (default:" ' "wheelhouse/")'),
+        help="Directory to store delocated wheels (default: %(default)r)",
         default="wheelhouse/",
     )
     p.add_argument(
@@ -79,7 +82,7 @@ wheel will abort processing of subsequent wheels.
             "Do not update the wheel filename tags and WHEEL info"
             " to match the repaired platform tag."
         ),
-        default=True,
+        default=True,  # default: UPDATE_TAGS=True
     )
     p.add_argument(
         "--strip",
@@ -101,10 +104,11 @@ wheel will abort processing of subsequent wheels.
         action="store_false",
         help=(
             "Do not copy libraries located in the site-packages directory from "
-            "other packages (update rpath only). The developer will need to ensure "
-            "that the skipped libraries are listed the package install dependencies."
+            "other packages. Just update the `rpath` only. The developer will "
+            "need to ensure that the skipped libraries are listed the package "
+            "install dependencies."
         ),
-        default=True,
+        default=True,  # default: COPY_SITE_LIBS=True
     )
     p.set_defaults(func=execute)
 
--- a/src/auditwheel/main_repair.py
+++ b/src/auditwheel/main_repair.py
@@ -95,6 +95,17 @@ wheel will abort processing of subsequent wheels.
         help="Do not check for higher policy compatibility",
         default=False,
     )
+    p.add_argument(
+        "--no-copy-site-libs",
+        dest="COPY_SITE_LIBS",
+        action="store_false",
+        help=(
+            "Do not copy libraries located in the site-packages directory from "
+            "other packages (update rpath only). The developer will need to ensure "
+            "that the skipped libraries are listed the package install dependencies."
+        ),
+        default=True,
+    )
     p.set_defaults(func=execute)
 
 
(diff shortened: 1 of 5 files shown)


mattip:
@mayeut any thoughts?

mattip:
Could a maintainer allow running the workflow for this? NumPy and SciPy would like to create an openblas wheel, and this is a necessary component to not vendor the shared object into the NumPy/SciPy wheels.

auvipy:
started

auvipy:
builds are red sadly

src/auditwheel/repair.py
```diff
@@ -225,13 +225,13 @@ def is_valid_rpath(rpath: str) -> bool:
...
         for rp in rpaths
     ]
+    if new_rpath is not None:
+        rpaths.append(new_rpath)
     # Remove duplicates while preserving ordering
     # Fake an OrderedSet using OrderedDict
-    rpath_set = OrderedDict([(old_rpath, "") for old_rpath in rpaths])
-    if rpath is not None:
-        rpath_set[rpath] = ""
+    rpaths = list(OrderedDict.fromkeys(rpaths))
```
auvipy, line 232: can we just use dict instead here?

XuehaiPan: Since Python 3.6, the `dict` keys are kept in insertion order. But I think explicitly using `OrderedDict` improves readability.

auvipy: ok for now

src/auditwheel/main_repair.py
```diff
@@ -95,6 +95,17 @@ def configure_parser(sub_parsers):
...
+        "--no-copy-site-libs",
+        dest="COPY_SITE_LIBS",
+        action="store_false",
+        help=(
+            "Do not copy libraries located in the site-packages directory from "
+            "other packages (update rpath only). The developer will need to ensure "
+            "that the skipped libraries are listed the package install dependencies."
+        ),
+        default=True,
+    )
```
mattip, line 108: I would
- align the option name with the environment variable name, `--no-copy-site-libs` and `COPY_SITE_LIBS` is confusing
- set the default to be the previous behaviour, so to always copy site libs by default.

mattip: I think the wrong default is what is causing tests to fail. The produced wheels in tests are not manylinux-compliant, so `auditwheel show` is setting the `linux_x86_64` tag rather than a `manylinux_2_17_x86_64` one.

mattip:
Tests are needed to show this works. Perhaps you could parameterize the failing tests in `test_manylinux.py`:
- test_build_wheel_depending_on_library_with_rpath`
- test_build_repair_multiple_top_level_modules_wheel`

src/auditwheel/main_repair.py
XuehaiPan: > align the option name with the environment variable name, `--no-copy-site-libs` and `COPY_SITE_LIBS` is confusing

I followed the design of `--no-update-tags` and `dest="UPDATE_TAGES"`:

https://github.com/pypa/auditwheel/blob/560509888bdb9cdfa98b41cc1641ec749b64b949/src/auditwheel/main_repair.py#L74-L83

Here `default=True` means `args.UPDATE_TAGES` is `True` if the option `--no-update-tags` is not present. If the option `--no-update-tags` present, `action="store_false"` changes `args.UPDATE_TAGES` to `False`.

> set the default to be the previous behaviour, so to always copy site libs by default.

`args.COPY_SITE_LIBS=True` is the default behavior.

mattip: > `args.COPY_SITE_LIBS=True` is the default behavior.

OK, sorry for the noise. Something is still off since the tests are failing.

XuehaiPan:
I changed the line in `assert_show_output`:

```diff
-assert actual_match, f"No match for tag {match['tag']}"
+assert actual_match, f"No match for tag {match['tag']}\nmatch.groupdict()\n:{match.groupdict()!r}output:\n{output!r}"
```

to investigate want happens in tests. I got:

XuehaiPan:
The tests passed for commit [`6f6e921` (#392)](https://github.com/pypa/auditwheel/pull/392/commits/6f6e9216f25107683a7d9a6cb15dd6556d2c1636): <https://github.com/pypa/auditwheel/runs/7854383503?check_suite_focus=true>.

But failed after commit: [`80b7152` (#392)](https://github.com/pypa/auditwheel/pull/392/commits/80b7152a140e295da92c133c8407c3bdf0425123)

XuehaiPan:
The tests failed for grouping duplicate libraries by real soname. I removed the grouping mechanism. Tests passed on my local machine. However, there may be duplicate libraries copied into the wheel file.

mattip:
It has been a while @mayeut can we get some eyes on this? It would help move openblas packaging as a wheel forward.

src/auditwheel/repair.py
```diff
@@ -79,16 +84,35 @@ def repair_wheel(
                         % soname
                     )
 
+                if not copy_site_libs and "site-packages" in str(src_path).split(
```
mayeut, line 87: There's no guarantee for the `site-packages` folder to have that name (in fact, it doesn't on debian python, it's name is `dist-packages`).

src/auditwheel/main_repair.py
```diff
@@ -95,6 +98,18 @@ def configure_parser(sub_parsers):
         help="Do not check for higher policy compatibility",
         default=False,
     )
+    p.add_argument(
+        "--no-copy-site-libs",
+        dest="COPY_SITE_LIBS",
+        action="store_false",
+        help=(
+            "Do not copy libraries located in the site-packages directory from "
+            "other packages. Just update the `rpath` only. The developer will "
```
mayeut, line 107: unfortunately, just updating the `rpath` is not enough.

mayeut:
As mentioned in https://github.com/pypa/auditwheel/issues/391#issuecomment-1272349511, it's not as easy as just updating the `rpath` because both packages might not be in the same `platlib` folder.

Given `rpath` is not an option, I think we're kind of in the same situation as #368 with the difference that the dependency is not an arbitrary shared object but one provided by another package. It's thus the responsibility of the package requiring the dependency to ensure proper loading of the shared object it depends on. It's not enough to declare the package as a dependency in metadata (& this must be done with care regarding ABI compatibility between different versions).

Given specific steps are required to ensure proper loading of shared object, I'm not sure we want to silently filter out all shared objects in `site-packages` and thus we're a bit more like in #368 were exceptions are explicitly passed in one-by-one.

@lkollar, any thoughts ?

mayeut:
Closing this PR in favor of #368 
Thanks for your efforts @XuehaiPan, despite the PR not getting merged, this allowed a broader discussion to get things moving forward I think.

PR #393, opened 2022-08-24
Improve log message in case of NonPlatformWheel error

CaselIT:
The current log message in case a wheel file contains no compiled extension is not very clear regarding that's the issue is.
I've tried to add a clarification that I how will make it clearer for uses to understand what when wrong

(I opened https://github.com/pypa/cibuildwheel/issues/1247 since it was not clear to me that the error meant, I though it was complaining about the file name abi identifier or something similar)

### DIFF
--- a/src/auditwheel/main_addtag.py
+++ b/src/auditwheel/main_addtag.py
@@ -24,16 +24,17 @@ def execute(args, p):
     import os
 
     from ._vendor.wheel.wheelfile import WHEEL_INFO_RE
-    from .wheel_abi import NonPlatformWheel, analyze_wheel_abi
+    from .wheel_abi import (
+        NOT_PLATFORM_WHEEL_MESSAGE,
+        NonPlatformWheel,
+        analyze_wheel_abi,
+    )
     from .wheeltools import InWheelCtx, WheelToolsError, add_platforms
 
     try:
         wheel_abi = analyze_wheel_abi(args.WHEEL_FILE)
     except NonPlatformWheel:
-        logger.info(
-            "This does not look like a platform wheel"
-            ", no compiled extension file found in the wheel archive"
-        )
+        logger.info(NOT_PLATFORM_WHEEL_MESSAGE)
         return 1
 
     parsed_fname = WHEEL_INFO_RE.search(basename(args.WHEEL_FILE))
--- a/src/auditwheel/main_addtag.py
+++ b/src/auditwheel/main_addtag.py
@@ -30,7 +30,7 @@ def execute(args, p):
     try:
         wheel_abi = analyze_wheel_abi(args.WHEEL_FILE)
     except NonPlatformWheel:
-        logger.info(NonPlatformWheel.LogMessage)
+        logger.info(NonPlatformWheel.LOG_MESSAGE)
         return 1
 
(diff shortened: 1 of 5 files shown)


CaselIT:
It seems that the fails in the pipeline are not connected to the changes in this PR, but are due to travis ci worker handing

src/auditwheel/main_addtag.py
```diff
@@ -30,7 +30,10 @@ def execute(args, p):
     try:
         wheel_abi = analyze_wheel_abi(args.WHEEL_FILE)
     except NonPlatformWheel:
-        logger.info("This does not look like a platform wheel")
+        logger.info(
+            "This does not look like a platform wheel"
+            ", no compiled extension file found in the wheel archive"
```
mayeut, line 35: auditwheel does not require the presence of a compiled extension. It requires the presence of at least 1 ELF file: either an executable or a shared library. A compiled extension is just a specific kind of shared library.

mayeut (changes requested):

src/auditwheel/main_addtag.py
CaselIT: I guess a compiled extension is the most common usage kind for cibuildwheel, but I agree that we should be more general.

Does the following work?

> , no executable or shared library found in the wheel archive

mayeut, line 35: Maybe something like the following: it has the precision about ELF file but also mention the extension bit as it might be more meaningful depending on one's knowledge of all of these details.

```suggestion
            ", no ELF executable or shared library file (including compiled Python C extension) found in the wheel archive"
```

PS: it probably won't pass linter validation as-is.

CaselIT: Thanks for the input. I'll update the change

CaselIT:
I've updated.

I've taken the liberty of creating a constant with the message, since it was used in 3 different places, but I'm not sure if where I placed it is the correct place. I can remove it if you prefer not having it

src/auditwheel/wheel_abi.py
```diff
@@ -51,6 +51,13 @@ class NonPlatformWheel(WheelAbiError):
     """No ELF binaries in the wheel"""
 
 
+NOT_PLATFORM_WHEEL_MESSAGE = (
```
mayeut, line 54: Can you make this a variable of the class `NonPlatformWheel` please ?

mayeut: > I've taken the liberty of creating a constant with the message, since it was used in 3 different places, but I'm not sure if where I placed it is the correct place. I can remove it if you prefer not having it

I almost suggested it to be done but didn't in the end. Since you've done it, I suggested another place which will also reduce the number of imports.

CaselIT: that was by second alternative. Will update

src/auditwheel/wheel_abi.py
```diff
@@ -50,6 +50,12 @@ class WheelAbiError(Exception):
 class NonPlatformWheel(WheelAbiError):
     """No ELF binaries in the wheel"""
 
+    LogMessage = (
```
mayeut, line 53:
```suggestion
    LOG_MESSAGE = (
```

mayeut (approved):

mayeut:
thanks @CaselIT !

PR #412, opened 2023-03-05
Fix CI workflows

mayeut:
Workflows are currently failing because of:
- `pre-commit` configuration
- wrong `gcc-toolset` used on `manylinux_2_28`

This PR updates both.


auvipy (approved):

auvipy (changes requested):
builds are failing

PR #419, opened 2023-04-10
Use armv7l policy for 64-bit arm kernel in 32-bit mode (armv8l)

mayeut:
Towards https://github.com/pypa/manylinux/pull/1455


mayeut:
@lkollar, I'm willing to take a chance on this one in order to get things started for armv7l builds on aarch64 systems with 32-bit userland. There are still things that require some checks but I think it will be easier to get feedback if we do release this (it could come with a big disclaimer that this is an alpha feature).

lkollar (approved):

auvipy (approved):

PR #434, opened 2023-06-21
Added '-t/--set-tags' option to 'addtag' subcommand

themarpe:
Main idea for this modification is to either "repair" armv7l wheels to armv6l, or to add both tags to the wheel.
Raspberry Pi usually report platform as armv7l but their toolchain defaults to armv6l compatible builds. If author of the wheel is sure that is the case, they may just make a copy/tweak to the built wheel to be marked as compatible for both.

Could also be spun off as a separate command.

Side observation, `add_platforms` does not act correct when a tag is specified as to be removed + to be added. In that case, remove takes precedence, but not in the filename. This makes the WHEEL file contain different tag specifiers than resulting wheel filename.

### DIFF
--- a/src/auditwheel/main_addtag.py
+++ b/src/auditwheel/main_addtag.py
@@ -19,6 +19,13 @@ def configure_parser(sub_parsers):
         type=abspath,
         default="wheelhouse/",
     )
+    p.add_argument(
+        "-t",
+        "--set-tags",
+        dest="TAGS",
+        help=("Set specified tags explicitly"),
+        nargs="+",
+    )
     p.set_defaults(func=execute)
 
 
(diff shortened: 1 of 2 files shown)


mayeut:
I think that wheel cli does that (& probably more).
I'd rather not have auditwheel reimplement the same thing.

@themarpe, can you confirm that [wheel tags](https://wheel.readthedocs.io/en/stable/reference/wheel_tags.html) fits your use case ?

themarpe:
Can verify this behavior is covered by `wheel` package already. Thanks @mayeut 

To add additional tag[s] using `wheel`, example: `linux_armv6l`:
```
python3 -m wheel tags --platform-tag +linux_armv6l [.whl]
```

PR #438, opened 2023-07-13
When repairing a pure python wheel, exit code should be 0

rdbisme:


rdbisme:
Based on https://github.com/pypa/auditwheel/issues/47, when doing `auditwheel repair` on a pure python wheel, the exit code should not be 1. 

My use case is that I'm running repair on all the wheel produced by some colleague, and if they're pure python, it makes the CI pipeline to fail.

mayeut:
I'd still want to fail on this one.
How about a specific error code for this case ?

cc @lkollar for another opinion

rdbisme:
> I'd still want to fail on this one. How about a specific error code for this case ?
> 
> cc @lkollar for another opinion

That will require special handling in the CI for my use case anyway. Why a `noop` should exit as a failure? I'd expect for it to not error if there's nothing to do...

PR #443, opened 2023-09-03
Repair ELF executables in the "scripts" directory

### LINKED ISSUE #340: Scripts are not correctly repaired
Binaries that go to the `.data/scripts` folder are not fixed correctly, the rpath value is wrong. There's no way to know the correct rpath value if we keep the executable in the scripts folder, as that depends on the Python installation.

The only solution I see is to move the scripts to platlib, where we know its relative path to the `.libs` folder, and place a wrapper in its place that calls `os.execv` on the actual binary.

Does anyone have other ideas?

See https://github.com/FFY00/mesonpy/issues/3#issuecomment-926841654

chriskuehl:
Fixes #340

This is an early draft and request for feedback. I'm happy to add tests but wanted to request an initial review first to see if this approach is workable.

## Summary

Some packages contain ELF executables in their "scripts" directory which link to shared libraries embedded in the wheel. Currently auditwheel can't correctly repair the RPATH of these executables, and repaired wheels will have broken executables when installed.

The approach here is to:
* Move executables into a new `{package}.scripts` directory in platlib, similar to the `{package}.scripts` directory.
* Replace the executable in the old `scripts` directory with a shim that execs the relocated executable when called.

This works because the relocated executables can have a consistent RPATH (`$ORIGIN/../{package}.libs`). This is required since installed executables in `scripts` don't have a consistent relative path to platlib (it varies based on how the wheel is installed).

This is roughly following the approach suggested in https://github.com/pypa/auditwheel/issues/340#issuecomment-932930526.

## Demonstration

I'm testing using a wheel I built for `uwsgi==2.0.22`:

```
$ auditwheel repair --plat manylinux_2_31_x86_64 uWSGI-2.0.22-cp39-cp39-linux_x86_64.whl  
INFO:auditwheel.main_repair:Repairing uWSGI-2.0.22-cp39-cp39-linux_x86_64.whl
INFO:auditwheel.wheeltools:Previous filename tags: linux_x86_64
INFO:auditwheel.wheeltools:New filename tags: manylinux_2_31_x86_64
INFO:auditwheel.wheeltools:Previous WHEEL info tags: cp39-cp39-linux_x86_64
INFO:auditwheel.wheeltools:New WHEEL info tags: cp39-cp39-manylinux_2_31_x86_64
INFO:auditwheel.main_repair:
Fixed-up wheel written to /home/ckuehl/proj/auditwheel/wheelhouse/uWSGI-2.0.22-cp39-cp39-manylinux_2_31_x86_64.whl
```

Installing the wheel produces a working `uwsgi` binary:

```
$ pip install ./uWSGI-2.0.22-cp39-cp39-manylinux_2_31_x86_64.whl 
Processing ./uWSGI-2.0.22-cp39-cp39-manylinux_2_31_x86_64.whl
Installing collected packages: uWSGI
Successfully installed uWSGI-2.0.22

$ which uwsgi
/tmp/tmp.VQ8OCc05xH/venv/bin/uwsgi

$ uwsgi --version
2.0.22

# Also works without the virtualenv activated.
$ deactivate
$ /tmp/tmp.VQ8OCc05xH/venv/bin/uwsgi --version
2.0.22
```

The `uwsgi` binary on my PATH was replaced with a wrapper script:
```
$ cat /tmp/tmp.VQ8OCc05xH/venv/bin/uwsgi
#!/tmp/tmp.VQ8OCc05xH/venv/bin/python
import os
import sys
import sysconfig


if __name__ == "__main__":
    os.execv(
        os.path.join(sysconfig.get_path("platlib"), 'uWSGI.scripts/uwsgi'),
        sys.argv,
    )
```

The real executable was relocated into site-packages:

```
$ ls -l /tmp/tmp.VQ8OCc05xH/venv/lib/python3.9/site-packages/uWSGI.scripts/uwsgi 
-rwxr-xr-x 1 ckuehl ckuehl 1.5M Sep  3 00:09 /tmp/tmp.VQ8OCc05xH/venv/lib/python3.9/site-packages/uWSGI.scripts/uwsgi
```

...and it has an RPATH to the libs directory:

```
$ readelf -d /tmp/tmp.VQ8OCc05xH/venv/lib/python3.9/site-packages/uWSGI.scripts/uwsgi | grep RPATH
 0x000000000000000f (RPATH)              Library rpath: [$ORIGIN/../uWSGI.libs]
```

## Comparison to current auditwheel

Running the same repair on the `main` branch produces a repaired wheel but the `uwsgi` binary doesn't work after installation:

```
$ uwsgi --version
uwsgi: error while loading shared libraries: libpcre-2092419b.so.3.13.3: cannot open shared object file: No such file or directory
```

...and the rewritten RPATH points to an invalid location:
```
$ readelf -d venv/bin/uwsgi | grep RPATH
 0x000000000000000f (RPATH)              Library rpath: [$ORIGIN/../../uWSGI.libs]
```


mayeut:
Thanks @chriskuehl, this looks good.

Could you add 2 tests for this:
- an executable without dependencies shall be left untouched at the original location (there's a penalty using a shim so if not required, might as well not use one)
- an executable with dependencies shall use the shim.

matejsp:
I have the same issue. Any chance to complete this PR?

chriskuehl:
> Could you add 2 tests for this:
> 
> * an executable without dependencies shall be left untouched at the original location (there's a penalty using a shim so if not required, might as well not use one)
> * an executable with dependencies shall use the shim.

I've added tests that cover this under the existing `test_build_wheel_with_binary_executable` test.

For reference, the updated test fails on `main` like this as expected (exhibiting the issue reported in #340):

```
INFO     test_manylinux:test_manylinux.py:151 docker exec 2d9783f8b742: ['/usr/local/bin/testprogram', '4']                                                                                                                                                                                                                  
/usr/local/bin/testprogram: error while loading shared libraries: libgsl-b64c76bc.so.23.1.0: cannot open shared object file: No such file or directory                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                             
FAILED
```

Let me know if you'd like me to move these assertions to an entirely new test case or to rearrange things in some other way.

tests/integration/test_manylinux.py
```diff
@@ -175,7 +175,7 @@ def tmp_docker_image(base, commands, setup_env={}):
 
     logger.info("Made image %s based on %s", image.short_id, base)
     try:
-        yield image.short_id
+        yield image.id
     finally:
```
chriskuehl, line 179: I left more details in the commit description, but this makes the tests work under Podman and seems relatively benign. Happy to revert if you prefer.

src/auditwheel/repair.py
```diff
@@ -232,3 +235,58 @@ def _resolve_rpath_tokens(rpath: str, lib_base_dir: str) -> str:
...
+    return """\
+#!python
+import os
+import sys
+import sysconfig
+
+
+if __name__ == "__main__":
+    os.execv(
+        os.path.join(sysconfig.get_path("platlib"), {binary_path!r}),
```
mayeut, line 287: I wonder if this works for any possible deployment scenarios.
Maybe using [importlib.metadata](https://docs.python.org/3/library/importlib.metadata.html#distribution-files) on python >= 3.8 might be a more robust way to get the real path.

Maybe this could be an improvement left for a later PR if an issue is reported.

mayeut (approved):
Thanks @chriskuehl !

This looks good. I left a comment but I don't think it's a blocker (unless @lkollar disagrees).

lkollar (approved):

PR #448, opened 2023-09-11
chore: rely on packaging to parse wheel filename

mayeut:
Python 3.12 removed the `distutils` module from stdlib.
Adherence to this module in auditwheel vendored (wheel) sources shall be removed.

Given updating wheel would be a longer task, switch to packaging to parse wheel filename in order to drop some wheel vendored files which are relying on `distutils`.


src/auditwheel/wheeltools.py
```diff
@@ -223,9 +224,8 @@ def add_platforms(
         out_dir = "."
         wheel_fname = basename(wheel_ctx.in_wheel)
 
-    parsed_fname = WHEEL_INFO_RE.match(wheel_fname)
-    fparts = parsed_fname.groupdict()
-    original_fname_tags = fparts["plat"].split(".")
+    wheel_components = parse_wheel_filename(wheel_fname)
```
lkollar, line 227: I would make this consistent with `main_addtag.py` and unpack the tuple in the assignment as well..

lkollar (approved):
Nice! This is also means that distros which devendor wheel won't have to do this anymore.

mayeut:
> This is also means that distros which devendor wheel won't have to do this anymore.

Not just yet, there's still a dependency on `wheel.pkginfo` for now (which has been removed from latest versions of wheel as such)

Issue #444, opened 2023-09-04
Addtag always tries to add linux_x86_64 tag


lkollar:
> This might be a bug in `addtag`, but I'm not even sure what this subcommand is useful for. It seems to try to assign the minimum compatible tag in the file name, but from my limited testing it always ends up with the `linux_x86_64` tag, which isn't useful. So if you pass a wheel which only has a `linux_x86_64` tag, it does nothing, and for wheels which have manylinux tags it also adds `linux_x86_64`, which isn't very useful either. I wonder if this is a regression, as there aren't any real tests for this functionality...

_Originally posted by @lkollar in https://github.com/pypa/auditwheel/issues/439#issuecomment-1705711191_

mayeut:
I wonder if we should simply remove this command in auditwheel 6.0.0 ?
As you mentioned, I don't know how useful this is...

lkollar:
I'd in favour of removing it. It doesn't seem useful in its current form at all.

PR #455, opened 2023-09-25
Remove the `addtag` subcommand

lkollar:
This subcommand has been broken for at least a few releases and doesn't provide any useful functionality. Removing it will at least avoid confusing users when they try to use it.

Closes: #444


auvipy (commented):
is there any consequence on removing the?

mayeut (approved):

PR #458, opened 2023-09-27
Demonstrate how ELF extraction/utilities can be done using SQL

fzakaria:
I'm working on [sqlelf](https://github.com/fzakaria/sqlelf) which is a tool that lets you explore ELF files using SQL.

What's really interesting and neat about this tool is that it lets you declaratively write what data you are interested in without knowing the internals of the ELF file format or working with the the necessary ELF parsing library.

The snippets are also very easily runnable outside the codebase which makes validating the tests separately super easy.

Here is a current _draft_ where I have replaced a good number of the functions with their respective SQL equivalent.
I'm interested in feedback or thoughts on the approach.

We can get quite fancy with the SQL to even split the strings of RUNPATH for instance for instance.
image

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -29,6 +29,7 @@ include_package_data = True
 install_requires =
     packaging>=20.9
     pyelftools>=0.24
+    sqlelf>=0.1.0
     importlib_metadata; python_version < "3.8"
 packages = find:
 package_dir =
--- a/setup.cfg
+++ b/setup.cfg
@@ -29,7 +29,7 @@ include_package_data = True
 install_requires =
     packaging>=20.9
     pyelftools>=0.24
-    sqlelf>=0.1.0
+    sqlelf>=0.2.0
     importlib_metadata; python_version < "3.8"
 packages = find:
 package_dir =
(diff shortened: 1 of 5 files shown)


lkollar:
This looks like a cool project.

> What's really interesting and neat about this tool is that it lets you declaratively write what data you are interested in without knowing the internals of the ELF file format or working with the the necessary ELF parsing library.

While this sounds nice, I think users still need to have some knowledge of ELF to be able to use sqlelf, and anyone contributing to auditwheel's parts dealing with ELF will need to have a good understanding of the format anyway. 

My main concern with this is that it would require developers to have fairly good SQL knowledge to contribute to the code base. While the equivalent Python code is often more verbose, I expect that anyone who's able to contribute to auditwheel will be able to fairly easily understand it, but this might not be the case with SQL code, which is a new language in this code base. And besides, mixing SQL with Python code means tools like linters, IDEs, etc. won't be able to help catch bugs, offer autocomplete, etc. 

So all in all, I'm not convinced that replacing pyelftools with sqlelf provides enough benefits to be worth doing.

mayeut:
Thanks for your interest in the project @fzakaria 
While indeed it looks like a cool project, I do agree with @lkollar w.r.t to the inclusion in auditwheel.

PR #466, opened 2023-11-06
Remove globals in policy module

lkollar:
I started working on this when I created #460 and ran into the problem with the globals in the `policy` module once again. These globals make it very difficult to properly test simple changes as the module-level globals get initialised on the first import which usually happens before any mocking can be done.

This is a large PR, but most of the changes are just moving code around.


src/auditwheel/policy/__init__.py
```diff
@@ -3,21 +3,107 @@
...
+}
+
+
+class WheelPolicies:
+    @staticmethod
+    def load():
+        libc_variant = get_libc()
+        policies_path = _POLICY_JSON_MAP[libc_variant]
+        policy_dict = json.loads(policies_path.read_text())
+        return WheelPolicies(policy_dict)
```
mayeut, line 38: is `load()` really needed ? couldn't this be done in `__init__()` ?

src/auditwheel/policy/__init__.py
```diff
@@ -3,21 +3,107 @@
...
+
+    def __init__(self, policies: dict) -> None:
```
mayeut, line 40: c.f. previous comment on `load()`

Maybe it would also be good to get arch/libc/musl as optional arguments in `__init__` but that could come later (I'm thinking about cross-platform repair / cross-platform check).

src/auditwheel/policy/__init__.py
```diff
@@ -181,14 +204,124 @@ def get_replace_platforms(name: str) -> list[str]:
...
-# These have to be imported here to avoid a circular import.
-from .external_references import lddtree_external_references  # noqa
-from .versioned_symbols import versioned_symbols_policy  # noqa
+def _load_policy_schema():
+    with open(join(dirname(abspath(__file__)), "policy-schema.json")) as f_:
+        schema = json.load(f_)
+    return schema
+
+
+def versioned_symbols_policy(
```
mayeut, line 213: should this become a member of WheelPolicies ?

src/auditwheel/policy/__init__.py
```diff
@@ -181,14 +204,124 @@ def get_replace_platforms(name: str) -> list[str]:
...
+            matching_policies.append(p["priority"])
+
+    if len(matching_policies) == 0:
+        # the base policy (generic linux) should always match
+        raise RuntimeError("Internal error")
+
+    return max(matching_policies)
+
+
+def lddtree_external_references(
```
mayeut, line 254: should this become a member of WheelPolicies ?

src/auditwheel/policy/__init__.py
jvolkman: > Maybe it would also be good to get arch/libc/musl as optional arguments in `__init__` but that could come later (I'm thinking about cross-platform repair / cross-platform check).

This would be great. My project does cross-platform repair and jumps through module loading [hoops](https://github.com/jvolkman/repairwheel/blob/main/src/repairwheel/linux/monkeypatch.py#L6) currently. Although having everything in a `WheelPolicies` class should already help with that patching.

lkollar: @mayeut Do you mind if we do this in a follow up?
@jvolkman If you do end up needing this functionality, would you be willing to make a PR for it?

mayeut: > Do you mind if we do this in a follow up?

I don't mind

PR #472, opened 2024-01-02
Added the option to skip setting library SONAMEs

pvallet:
I am adding this option locally as I ended up needing it to circumvent this problem I encountered, so I figured I might as well create a PR:
https://stackoverflow.com/questions/77733250/using-patchelf-to-change-the-soname-of-libgldispatch-so-breaks-its-functionality

While not the common flow, it does seem that being able to skip setting the SONAMEs and only relying on file names can have its uses in some cases.

### DIFF
--- a/src/auditwheel/main_repair.py
+++ b/src/auditwheel/main_repair.py
@@ -80,7 +80,7 @@ wheel will abort processing of subsequent wheels.
     )
     p.add_argument(
         "--no-set-soname",
-        "SET_SONAME",
+        dest="SET_SONAME",
         action="store_false",
         help="Move and rename libraries but don't set their SONAME",
         default=True,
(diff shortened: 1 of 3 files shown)


pvallet:
Hello! Is there any way I could help push this forward? Should I create some tests for the new option? I hadn’t made them as the option I was taking inspiration from (no update tags) didn’t have any, but I could definitely add some

mayeut:
I'm not almost sure that we don't want to see this happening.
Copying without renaming/changing the SONAME is likely to break things in different environments.
While `libOpenGL.so` can't be whitelisted, I think it falls in the category where it can be excluded from being grafted altogether: https://github.com/pypa/auditwheel/pull/368 

cc @lkollar for another opinion on the matter.

HinTak:
FWIW, think this pull is likely barking up the wrong tree - the explanation is probably along https://github.com/pypa/cibuildwheel/issues/2118 - libGLdispatch is used between libA and libA_vendor,  where libA is one of GL family of libraries. You need to load them from system locations to find the hidden libA_vendor.

mayeut:
The 2 last comments point to using `--exclude`.

Issue #409, opened 2023-01-26
[Question] Reasons to mangle and hash .so names after copy


AlexanderSerov:
Hello. We had demand to supply shared libraries inside the wheel and while we do so encountered problem with symlinks in wheel. This is how we reach this [thread](https://discuss.python.org/t/symbolic-links-in-wheels/1945) and particularly [this](https://discuss.python.org/t/symbolic-links-in-wheels/1945/6) message. So we extensively rely on auditwheel api to implement our own custom logic to copy libraries.

We encountered problem with mangling though. Our production library relying on cuda libs. Before we start use auditwheel, this cuda libs were stored in file system, outside of package, and using custom environment variables we manage to add this cuda libs to the linker search path. The benefits of this approach - we can reuse this cuda libs to run other cuda dependent packages like pytorch or something else. As soon we started to copy libraries and mangle them we loose this ability.

In the [message](https://discuss.python.org/t/symbolic-links-in-wheels/1945/6) mentioned above, there is brief mention "_IMO if you want to ship shared libraries in a wheel, then you should take the search problem seriously, and not rely on the linker’s naming scheme for system libraries._". What I want to ask is expand this abstract warning with concreate examples or some issues already raised or some articles on this subject... As far I i know, on linux for example, the linker search libraries in following order: 1) look in rpath, 2) if libs missed in rpath, we go to look in system. As far as we control rpath, looks like not any randomness introduced in matter of library search paths.
The implication of design with mangled library names is each package use they own shared libraries - is not good for the RAM. If mangling not take place we can reuse them, again, using custom environment variables. So question is, Is recommendation to mangle .so name is strong enough to follow him?

AlexanderSerov:
@njsmith

AlexanderSerov:
ChatGPT agreed with me
![image](https://user-images.githubusercontent.com/33323782/220898916-4dc4210c-105f-4ac4-956d-360caba6ca87.png)

njsmith:
There are two challenges:

- There's no guarantee that all your packages will be installed into a single `site-packages/` directory. There are lots of ways to set up a Python environment, and all you're really guaranteed if you depend on `foo` is that `import foo` will find the `foo` package -- maybe via `$PYTHONPATH`, maybe via some exotic `sys.metapath` trick, who knows. So there's no way for package A to reliably point an `RPATH` at package B, because package A doesn't know where its files will be located relative to package B.

[...]

PR #473, opened 2024-01-04
feat:add --no-update-soname flag

vimiix:
An implementation for this issue:

https://github.com/pypa/auditwheel/issues/409#issuecomment-1875104260


mayeut:
duplicates - in some form - #472

PR #474, opened 2024-01-07
fix: `--exclude libfoo.so` shall ignore dependencies of `libfoo.so`

mayeut:
When using `--exclude libfoo.so`, dependencies of `libfoo.so` are still being analyzed & grafted.
This commit moves the exclusion analysis to `lddtree` and filters  `libfoo.so` `DT_NEEDED` entries thus excluding its dependencies from the tree.

### DIFF
--- a/src/auditwheel/lddtree.py
+++ b/src/auditwheel/lddtree.py
@@ -300,6 +300,7 @@ def lddtree(
     prefix: str = "",
     ldpaths: dict[str, list[str]] | None = None,
     display: str | None = None,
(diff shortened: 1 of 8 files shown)


lkollar (approved):
LGTM. The one thing I would add is that you may want to make the `exclude` argument optional in `get_wheel_elfdata` and `analyze_wheel_abi` as it isn't always necessary and otherwise they require an empty set to be passed in (i.e. in tests). Not a big deal though.

Issue #451, opened 2023-09-14
missing RPATH in secondary SO packed into the wheel


mattip:
Maybe related to #363 and others.

I may have something similar in a wheel build of OpenBLAS. It packages a OpenBLAS shared object, which depends on libgfortran, which depends on libquadmath. Both of the support so's are packed by auditwheel into the wheel, but the RPATH in libgfortran is not set to `$ORIGIN`, so libquadmath is not found. I uploaded the auditwheel-repaired wheel to testpypi so the issue can be reproduced:
```
$ python3.11 -m venv /tmp/test
$ source /tmp/test/bin/activate
$ pip install -i https://test.pypi.org/simple/ scipy-openblas32==0.3.23.293

# This works, `libopenblas_python` can find its dependencies:
$ python -m scipy_openblas32
OpenBLAS using 'b'OpenBLAS 0.3.23.dev DYNAMIC_ARCH NO_AFFINITY Zen MAX_THREADS=64''

# But ldd shows a problem when looking at libgfortran, note the "not found"
$ ldd /tmp/test/lib/python3.11/site-packages/scipy_openblas32/lib/libopenblas_python.so 
	linux-vdso.so.1 (0x00007ffd9fa9e000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007ff314c90000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007ff314c8b000)
	libgfortran-040039e1.so.5.0.0 => /tmp/test/lib/python3.11/site-packages/scipy_openblas32/lib/./libgfortran-040039e1.so.5.0.0 (0x00007ff314600000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007ff3143d8000)
	/lib64/ld-linux-x86-64.so.2 (0x00007ff316e3d000)
	libquadmath-96973f99.so.0.0.0 => /tmp/test/lib/python3.11/site-packages/scipy_openblas32/lib/./libquadmath-96973f99.so.0.0.0 (0x00007ff314000000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007ff314c6d000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007ff314c4d000)
[...]

PR #477, opened 2024-02-02
Set RPATH on copied libs to the folder where they're copied ($ORIGIN)

alopezz:
I believe this would address #451.

We use a tool that uses `ldd` to make sure software packages don't have outside dependencies. After starting to try using auditwheel, `.so` files under the libs folder are flagged because the RPATH is set to `dest_dir`, which is not right. Even though it should never have runtime implications (because the RPATH of the first library in the dependency chain is correct), we'd rather fix this upstream rather than having to add explicit allowlisting to our setup.

This fixes that by setting RPATH to `$ORIGIN` in those cases. I've done some manual testing and that seems to help `ldd` find the dependencies properly. That looks like the right thing, but I may very well be missing something.

I'm happy to add tests if required (pointers regarding to where to look at would be appreciated in that case).

### DIFF
--- a/src/auditwheel/repair.py
+++ b/src/auditwheel/repair.py
@@ -161,7 +161,7 @@ def copylib(src_path: str, dest_dir: str, patcher: ElfPatcher) -> tuple[str, str
     patcher.set_soname(dest_path, new_soname)
 
     if any(itertools.chain(rpaths["rpaths"], rpaths["runpaths"])):
-        patcher.set_rpath(dest_path, dest_dir)
+        patcher.set_rpath(dest_path, "$ORIGIN")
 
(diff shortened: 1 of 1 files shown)


mayeut:
> I'm happy to add tests if required

Tests shall already be in place for this. They are probably broken if there's no test failure... I'll take a look.

alopezz:
@mayeut Is there any reason to not set the RPATH always to `$ORIGIN` instead of only replacing it when there's an RPATH already?
https://github.com/pypa/auditwheel/blob/db7962149a3e971eba3974a58edcb2226c787f8d/src/auditwheel/repair.py#L166-L167

Issue #488, opened 2024-06-01
repair add empty `${package}.lib` directory to wheel


trim21:
I have a wheel build from manylinux image and it doesn't need extra library.

But after `auditwheel repair`, a empty `${package}.libs` directory is added to wheel.

Would be nice auditwheel doesn't add a empty directory

PR #489, opened 2024-06-01
fix: don't create empty .libs directory

trim21:
close https://github.com/pypa/auditwheel/issues/488

only create `${package}.libs` when needed

### DIFF
--- a/tests/integration/test_manylinux.py
+++ b/tests/integration/test_manylinux.py
@@ -1004,7 +1004,9 @@ class TestManylinux(Anylinux):
 
         with zipfile.ZipFile(os.path.join(io_folder, repaired_wheel)) as z:
             for file in z.namelist():
-                assert not file.startswith("testsimple.libs"), "should not have empty .libs folder"
+                assert not file.startswith(
+                    "testsimple.libs"
+                ), "should not have empty .libs folder"
 
         docker_exec(docker_python, f"pip install /io/{repaired_wheel}")
         docker_exec(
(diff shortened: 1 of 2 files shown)


mayeut:
Thanks for your contribution, would you be able to add a test case for this as well (in `tests/integration/test_manylinux.py:AnyLinux`, the `testsimple` project can be used for the new test as it has no external dependency) ?

trim21:
> Thanks for your contribution, would you be able to add a test case for this as well (in `tests/integration/test_manylinux.py:AnyLinux`, the `testsimple` project can be used for the new test as it has no external dependency) ?

where is `AnyLinux`? didn't find this with word search

tests/integration/test_manylinux.py
```diff
@@ -1000,8 +999,14 @@ def test_build_wheel_compat(
             repaired_tag = f"{expect_tag}.{target_tag}"
         repaired_wheel = f"testsimple-0.0.1-{PYTHON_ABI}-{repaired_tag}.whl"
         assert repaired_wheel in filenames
+
         assert_show_output(manylinux_ctr, repaired_wheel, expect, True)
 
+        with zipfile.ZipFile(os.path.join(io_folder, repaired_wheel)) as z:
+            for file in z.namelist():
+                if file.startswith("testsimple.libs"):
+                    raise Exception("should not have empty .libs folder")
```
mayeut, line 1008:
```suggestion
                assert not file.startswith("testsimple.libs"), "should not have empty .libs folder"
```

trim21: thanks, forget there is an assert statment...

auvipy (approved):

PR #502, opened 2024-07-30
Add support for an extra library name tag

sgillies:
Is prepended to the shorthash, with a dash separator. I think my weird Rasterio and Fiona wheels are the only things that will ever benefit from this.

Also apologies for mixing formatting with the new feature. I ran `darker` in my fork. If that's unacceptable, I understand.


sgillies:
Ugh, sorry about the noise here. GitHub is letting us down.

PR #503, opened 2024-07-30
Add option to add an extra lib name tag to wheel's short hash

sgillies:
I've been using this on wheels for my Fiona and Rasterio projects, which may be used together and which have a score of potentially overlapping dependencies. If it's not deemed useful to auditwheel, I understand.

### DIFF
--- a/src/auditwheel/main_repair.py
+++ b/src/auditwheel/main_repair.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import argparse
 import logging
+import os
 from os.path import abspath, basename, exists, isfile
 
 from auditwheel.patcher import Patchelf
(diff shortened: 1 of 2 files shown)


mayeut:
I don't understand what problem gets solved by this. Can you explain the issue you're trying to solve ?
My understanding is that overlapping dependencies is not an issue given the hash is added and thus, whatever version gets loaded, it'll work as they are the same version.

sgillies:
@mayeut my problem is that the Rasterio https://pypi.org/project/rasterio/ and Fiona https://pypi.org/project/fiona wheels, which I maintain, and are widely used in the GIS field, carry some of the same libraries and are sometimes used together. Some of the cornerstone GIS libraries, like GEOS and GDAL, are particular about being loaded twice. I've found that adding extra bits to the shared library names in addition to the short hash (which is likely the same for both rasterio and fiona) helps disambguiate the libraries and prevent some conflicts.

It's a pretty special problem, I admit.

PR #507, opened 2024-08-06
Make order of files in repaired wheel deterministic

bemoody:
Currently, when running `auditwheel repair`, the contents of the output whl file are unpredictable:
- The order of entries in the zip archive is unpredictable.
- The order of lines in the "RECORD" file is unpredictable.
In both cases, the order is dependent on the order of entries returned by `os.walk`.

This is a problem for build reproducibility - provided that the build process is sufficiently well defined, different people should be able to run the same process on different machines and get identical outputs.

Note that when `setuptools` or `wheel` generates a whl file, it does something similar (see `WheelFile.write_files` in `wheel.wheelfile`.)  The code here won't do quite the same as what setuptools does, but that shouldn't be a problem.

### DIFF
--- a/src/auditwheel/tools.py
+++ b/src/auditwheel/tools.py
@@ -77,6 +77,11 @@ def dir2zip(in_dir: str, zip_fname: str, date_time: datetime | None = None) -> N
         for root, dirs, files in os.walk(in_dir):
             dirs.sort()
             files.sort()
+            if root == in_dir:
+                # Place the contents of *.dist-info at the end of
+                # the archive, as recommended by PEP 427.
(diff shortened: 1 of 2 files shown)


mayeut (changes requested):
Thanks for the PR

According to https://peps.python.org/pep-0427/#recommended-archiver-features, it is recommended to place the .dist-info folder at the end of the archive.

If we're ensuring the order for build reproducibility, can this be taken into account please ?

bemoody:
Sure, that makes sense and is easy to do.

mayeut:
fixed by #517

Issue #500, opened 2024-07-22
Allow wildcards in `--exclude`


KyleFromNVIDIA:
It would be nice if the `--exclude` argument supported wildcards, like so:

```
auditwheel repair --exclude 'libexternallib.so.*' my_wheel.whl
```

petermbauer:
I would like to prevent any libraries from being added, hope this can be achieved then with `--exclude "*"`.

KyleFromNVIDIA:
> I would like to prevent any libraries from being added, hope this can be achieved then with `--exclude "*"`.

I just added a test to #508 to confirm that it supports this.

PR #508, opened 2024-08-06
Add ability to pass wildcard arguments to `--exclude`

KyleFromNVIDIA:
Fixes https://github.com/pypa/auditwheel/issues/500


ethanhs-nv:
Could you add a test verifying that sequences work? e.g. `libfoo.[1,2]` matches `libfoo.1`/`libfoo.2`, but not `libfoo.3`?

oraluben:
Hi @mayeut, I’d like to draw your attention to this PR for it can support wildcard `exclude`, which can support to exclude specific libs without knowing the detailed filename (e.g. to exclude `/usr/local/cuda*`). This feature can reduce a large amount of my work, please let me know if you think it's a good idea to merge this so I can try to find some other way.

Thanks!

oraluben (approved):

nickjbrowning:
I also have a similar issue, where for example I need to exclude: `libcudart-8774224f.so` and `libnvToolsExt-847d78f2.so` from my wheels. Wildcards would certainly help here for building wheels across different docker containers.

KyleFromNVIDIA:
> Could you add a test verifying that sequences work? e.g. libfoo.[1,2] matches libfoo.1/libfoo.2, but not libfoo.3?

@emmatyping-nv I had previously abandoned this PR because #411 had given me the impression that the PyPA team didn't want this change, but since there now appears to be activity on it from the maintainers, I went ahead and added this test.

mayeut:
> I had previously abandoned this PR because https://github.com/pypa/auditwheel/pull/411 had given me the impression that the PyPA team didn't want this change, but since there now appears to be activity on it from the maintainers, I went ahead and added this test.

I won't have time to re-review this properly before the end of the year, maybe @lkollar will ?
Given the high demand for this feature, we should find a middle-ground to allow wildcard while emphasizing that it should be used with caution (c.f. the comment in #411 where the feature would allow not to reflect dependencies properly in metadata).
One way would be to have a toggle to enable fnmatch with its help mentioning that proper metadata is the packager responsibility, possibly redirecting to #411.

mayeut:
I only added a comment in the help message in the end. Thanks @KyleFromNVIDIA

rdbisme:
Thanks for resurrecting my PR and finalize it :)

Issue #244, opened 2020-05-30
auditwheel repair --plat manylinux2014_aarch64


JonasVautherin:
It seems to me that `manylinux2014_aarch64` support was added to auditwheel (see [this PR](https://github.com/pypa/auditwheel/pull/202/files#)), but when I run:

```
auditwheel repair --plat manylinux2014_aarch64 my-manylinux2014-aarch64.whl
```

I get:

```
auditwheel repair: error: argument --plat: invalid choice: 'manylinux2014_aarch64' (choose from 'linux_x86_64', 'manylinux1_x86_64', 'manylinux2010_x86_64', 'manylinux2014_x86_64')
```

I am on auditwheel 3.1.1, python 3.8. Is that supported but rejected by the CLI, or is it not supported yet?

JonasVautherin:
Looking into the codebase, I see it is triggered [here](https://github.com/pypa/auditwheel/blob/8fc14ac8943fd1cd8f6b350ce6f0432f749ef17c/auditwheel/tools.py#L126).

And the choices come from `**kwargs`, here:

```python
super().__init__(
            default=default,
            required=required,
            **kwargs
        )
```

Unfortunately, I don't really understand where that comes from in the first place. So to answer part of my question: it seems that at least the CLI is refusing the platform. It doesn't say whether or not `manylinux2014-aarch64` is supported (should it pass this step), though.

lkollar:
Are you running this on an ARM 64 machine? From the available choices above, it seems this was run on an x86 box. You can only repair wheels on a host with matching architecture.

JonasVautherin:
Oh, that makes sense now that you say it. I am cross-compiling, from x86_64.

This said, do I actually need to repair a wheel built with manylinux? Or is it fine to just rename it? I tried it on test.pypi.org by just renaming and it seemed fine... In other words, how relevant is auditwheel for manylinux builds?

mayeut:
> This said, do I actually need to repair a wheel built with manylinux? Or is it fine to just rename it? I tried it on test.pypi.org by just renaming and it seemed fine...

For very simple extensions, auditwheel will not change much (but will still do checks before hand):
- `Tag` in file `{distribution}-{version}.dist-info/WHEEL` e.g. `cp37-cp37m-manylinux2010_x86_64`
- Name of the wheel

PyPI only checks the filename (for now). I don't know which checks installers are doing.
Standards implies both, what's actually implemented might be different (and subject to change to enforce standards a bit more).

[...]

PR #512, opened 2024-08-23
Add ability to repair wheels for other architectures

rhelmot:
#244 represents a problem in need of a solution - how to audit wheels for other architectures, e.g. those built with a cross toolchain? Not all build environments can set up binfmt_misc or are performant enough to build in a VM.

Here is a proposed solution: manually specifying the architecture for the wheels to patch. I have verified that this is able to repair wheels built for aarch64 using a cross toolchain on an x86_64 machine.


mayeut:
I think the arch shall be deduced from the input wheel. The information is already there, there should  be no need to ask the user for input.

mayeut:
Thanks @rhelmot for the deep analysis and the work done on this. I've not forgotten about it.
Indeed the wheel policies is really insistent on knowing everything before hand.
I want to get some work done before getting back to this but taking into account the work you've done.

PR #517, opened 2024-11-10
Walk directories in sorted order for reproducibility

dalcinl:
Implement a new utility function wrapping `os.walk` with the following modifications:
* recurse directories in sorted order
* recurse top-level *.dist-info/ directories last
* list filenames in sorted order
* list top-level *.dist-info/RECORD files last

### DIFF
--- a/src/auditwheel/tools.py
+++ b/src/auditwheel/tools.py
@@ -5,7 +5,7 @@ import os
 import subprocess
 import zipfile
 from datetime import datetime, timezone
-from typing import Any, Iterable
+from typing import Any, Generator, Iterable
 
 
 def unique_by_index(sequence: Iterable[Any]) -> list[Any]:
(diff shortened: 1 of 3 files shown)


dalcinl:
There is an additional nit that would be trivial to address in this PR. Currently, auditwheel outputs zip files using the usual convention of storing directories as an entry of zero bytes with name ending in "/". This is exactly what the `zip` utility does in Linux, but it is not what the `wheel` package does (and causes trouble with namespace packages). 

What we could do (with a one-line change in this PR) is to still store subdirectories, but only if they are empty. Again, this is not what `zip` or `python -m zipfile` do. However, it is closer to what `wheel` does, albeit without the issue related to namespace packages. There is a minor backward incompatibility  in the sense that the output zip files would no longer have all directory entries, but only those directories that are empty. However, upon unzipping, there would be no difference in the output directory tree. BTW, the wheel file spec says nothing about empty directories.

dalcinl:
This PR covers #507 and provides additional features (.dist-info directories last, .dist-info/RECORD files last)

src/auditwheel/tools.py
```diff
@@ -69,15 +113,16 @@ def dir2zip(in_dir: str, zip_fname: str, date_time: datetime | None = None) -> N
...
         st = os.stat(in_dir)
         date_time = datetime.fromtimestamp(st.st_mtime, tz=timezone.utc)
     date_time_args = date_time.timetuple()[:6]
     compression = zipfile.ZIP_DEFLATED
     with zipfile.ZipFile(zip_fname, "w", compression=compression) as z:
-        for root, dirs, files in os.walk(in_dir):
-            for dir in dirs:
-                dname = os.path.join(root, dir)
+        for root, dirs, files in walk(in_dir):
+            if root != in_dir:
```
dalcinl, line 124: The following trivial change would skip adding zip file entries for non-empty directories. Empty directories are still added, as that may be important for the case of special wheels with data files packaging core non-Python libraries and executables (e.g. a wheel with a tool  that expects a `$venv_prefix/etc/<toolname>` to exist).

```suggestion
            if root != in_dir and not (dirs or files):
```

mayeut:
Removing entries for non-empty folders also removes the permissions attribute for those. I'm gonna revert this for now.

dalcinl:
@mayeut Maybe you want to keep the test for empty folders?

mayeut:
thanks @dalcinl

PR #538, opened 2025-02-09
chore(tests): build numpy with openblas on manylinux_2_34

mayeut:


mayeut:
will be handled by #535

PR #540, opened 2025-02-10
chore: better typing hints

mayeut:
This uncovered a few issues so might as well add some more checks.

### DIFF
--- a/src/auditwheel/policy/__init__.py
+++ b/src/auditwheel/policy/__init__.py
@@ -51,7 +51,7 @@ class WheelPolicies:
         if arch is None:
             arch = Architecture.get_native_architecture()
         policies = json.loads(_POLICY_JSON_MAP[libc].read_text())
-        self._policies = []
+        self._policies: list[dict[str, Any]] = []
         self._architecture = arch
         self._libc_variant = libc
         self._musl_policy = musl_policy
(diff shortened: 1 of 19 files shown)


src/auditwheel/policy/__init__.py
```diff
@@ -84,18 +84,18 @@ def architecture(self) -> Architecture:
...
     @property
-    def policies(self):
+    def policies(self):  # type: ignore[no-untyped-def]
         return self._policies
 
     @property
-    def priority_highest(self):
-        return max(p["priority"] for p in self._policies)
+    def priority_highest(self) -> int:
+        return max(int(p["priority"]) for p in self._policies)
```
auvipy, line 92: are this changes fully test covered? or the integration tests are enough for verifying them?

mayeut: both unit tests and integration tests are covering those: https://app.codecov.io/gh/pypa/auditwheel/pull/540/blob/src/auditwheel/policy/__init__.py

mayeut: One of the next todo will be to get some proper types for policies instead of dicts of Any which are a pain to work with in the first place.

auvipy: do we have any open issue for the TODO? I am accepting this PR

auvipy (approved):

PR #542, opened 2025-02-15
fix: manylinux_2_39 policy

mayeut:
Use latest update from https://github.com/mayeut/pep600_compliance
Ubuntu 24.04 added some libstdc++ symbols in its latest updates.

We're in a gray area here. When it was released, Ubuntu 24.04 did not have those symbols for libstdc++ however, the latest updates / security updates (since mid-october 2024) do have those symbols.


auvipy (changes requested):
may be this need a rebase now?

PR #543, opened 2025-02-15
chore: use a dataclass instead of a dict for policies

mayeut:
This reduces a bit more dict usage and gets a more straightforward Policy class.


auvipy (approved):
nice changes

PR #545, opened 2025-02-19
WIP: Improve audit time

oraluben:
Motivation: We're using auditwheel to repair our internal PyTorch build, the overall time is ~10m. We're attempting to reduce the redundant operations that we've found, e.g. unzips.

This PR is not finalized yet but reviews are welcome!

- [x] add log for time-consuming parts
- [x] do not re-unzip wheel if possible
- [x] multiprocess grafting libs
- [x] set default compress level to 4
- [ ] tests
- [ ] cleanup

Update:
repairing a torch wheel takes longer than building one (still ongoing):
image

PR #547, opened 2025-02-23
feat: add option to set zip compression level

oraluben:
Add `--zip-compression-level` and `AUDITWHEEL_ZIP_COMPRESSION_LEVEL` env to specify zip compression level.

This PR do not change the default level.

Part of #545.
As in https://github.com/pypa/auditwheel/pull/545#issuecomment-2676697741, we don't want to alter the default compress level. However I do want to add this argument to have the ability to change it.

I'm not very sure about the best place to add test for this.

### DIFF
--- a/src/auditwheel/tools.py
+++ b/src/auditwheel/tools.py
@@ -183,7 +183,20 @@ class EnvironmentDefault(argparse.Action):
         self.env = env
         if self.env_default:
             if type:
-                self.env_default = type(self.env_default)
+                try:
+                    self.env_default = type(self.env_default)
+                except Exception:
+                    self.option_strings = kwargs["option_strings"]
+                    args = {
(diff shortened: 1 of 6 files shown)


src/auditwheel/tools.py
```diff
@@ -157,11 +176,14 @@ def __init__(
...
         default: str | None = None,
         choices: Iterable[str] | None = None,
+        type: type | None = None,
         **kwargs: Any,
     ) -> None:
         self.env_default = os.environ.get(env)
         self.env = env
         if self.env_default:
+            if type:
+                self.env_default = type(self.env_default)
```
mayeut, line 186: can a test be added in tests/unit/test_tools.py in order to cover this line please ?

mayeut:
Thanks @oraluben

PR #548, opened 2025-02-23
feat: detect architecture/libc from wheel

### LINKED ISSUE #390: Automatically repair to the most compatible/oldest platform tag
Hello all,

I'm wondering if I'm thinking about this correctly; wouldn't I want my wheels to have the lowest manylinux, such that more users can install it?

I'm building wheels using the manylinux_2_28 environment. Some of the C++ code uses newer symbols than others and need a more constrained/newer manylinux platform.

As an example, I build both `foo.whl` and `bar.whl` in manylinux_2_28;
* `foo.whl` can be repaired to `manylinux_2_24_x86_64`
* `bar.whl` can be repaired to `manylinux_2_17_x86_64`

`auditwheel show` will give you this information, by printing it with a bunch of text.

Foo:
```
$ auditwheel show foo.whl

foo.whl
is consistent with the following platform tag:
"manylinux_2_24_x86_64".

The wheel references external versioned symbols in these
...
```
Bar:
```
$ auditwheel show bar.whl

bar.whl
is consistent with the following platform tag:
"manylinux_2_17_x86_64".

The wheel references external versioned symbols in these
```

Next, I want to run `auditwheel repair --plat=${most-compatible-platform} *.whl`

1. I can either try to parse the output of `auditwheel show` (using grep etc.) - little bit tricky, especially when using cibuildwheel
```
CIBW_REPAIR_COMMAND_LINUX=auditwheel repair --plat=$(auditwheel show {wheel} | grep | awk | whatever) {wheel}
```
2. Maybe `auditwheel show` can have a shorter output flag, e.g. `auditwheel show --only-plat`, so it can be more easily used:
```
$ auditwheel show --only-plat bar.whl
manylinux_2_17_x86_64
```
More usable in a compound bash command:
```
CIBW_REPAIR_COMMAND_LINUX=auditwheel repair --plat=$(auditwheel show --only-plat {wheel}) {wheel}
```
3. Or, perhaps an `auditwheel repair --auto-plat` to combine the logic

Perhaps I'm missing some knowledge as well:
* Is it even desirable to have this behavior?
* Is there a downside I'm not considering?
* Can this already be achieved with auditwheel by some feature or flag I missed?

Thanks in advance.

mayeut:
towards #512

Getting the musllinux version for cross-repaired musl wheels is not handled yet here.

fixes #390: the default platform is changed from highest policy to `auto` which represents the highest compatible platform tag for any given wheel.


mayeut:
the tests added in #551 found an issue with auto-detection on armv7l so this will need to be corrected.

auvipy (approved):
this looks good to me

PR #554, opened 2025-03-07
feat: use thread pool

oraluben:
Use multithread to run IO-bounded jobs (e.g. shell command) concurrently.

Part of #545 

TODO: 
- [x] pass test with j=1
- [x] fuzz with different j
- [x] cli argument
- [x] doc
- [ ] test
- [ ] test time


oraluben:
Mark as ready-for-review for it passes all tests.

mayeut:
@oraluben, do you have any timings for this one ? I'd rather not increase the complexity of auditwheel if the improvement is small given maintainer bandwidth.

PR #559, opened 2025-03-16
chore: use pyproject.toml instead of setup.py/setup.cfg

mayeut:
Use dependency groups instead of extras for test/coverage/dev dependencies.
This allows to drop setup.py and move to a full description in pyproject.toml


setup.py
```diff
@@ -1,12 +0,0 @@
-from __future__ import annotations
```
auvipy, line 1: don't we need anything here? anymore?

mayeut: no, all is in pyproject.toml now

PR #575, opened 2025-04-16
lddtree: support alternative LD_LIBRARY_PATH called AUDITWHEEL_LD_LIBRARY_PATH

xnox:
Sometimes wheels are different only just by shared libraries that are
vendored in. Sometimes the wheel being repaired and the libraries that
need to be vendored in are accessible on the host that was not used
for building the wheel. In such cases, it could be the case that
shared libraries used by python that is executing the auditwheel
script, are incompatible (too old or too new) than those that need to
be vendored.

In such cases, LD_LIBRARY_PATH is not convenient to use, as the python
interpreter for the auditwheel script is crashing.

Add an AUDITWHEEL_LD_LIBRARY_PATH whichis used by lddtree, but
otherwise does not affect the python interpreter executing auditwheel
script.

This helps vendoring in older libraries from an alternative path, into
a wheel built against an older python, whilst otherwise running
auditwheel repair on a modern system with a much newer python. This is
particularly useful when stripping a wheel of vendored libraries, and
repair it again to update the shared library dependencies coming from
an unpacked chroot.

Separately it also helps creating alternative wheels with shared
libraries rebuilt with higher march settings. Given that python
upstream doesn't support loading optimised libraries as has been
implemented and supported in Intel ClearLinux Python which allows one
to have alternative march setting shared library deps.

### DIFF
--- a/src/auditwheel/lddtree.py
+++ b/src/auditwheel/lddtree.py
@@ -318,7 +318,7 @@ def load_ld_paths(
     ldpaths: dict[str, list[str]] = {"conf": [], "env": [], "interp": []}
 
     # Load up $LD_LIBRARY_PATH.
-    env_ldpath = os.environ.get("LD_LIBRARY_PATH")
+    env_ldpath = os.environ.get("AUDITWHEEL_LD_LIBRARY_PATH", default=os.environ.get("LD_LIBRARY_PATH"))
     if env_ldpath is not None:
         if root != "/":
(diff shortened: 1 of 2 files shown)


xnox:
I was unable to run test suite locally as i neither have old enough nor new enough pythons to run the tests, thus CI might fail, in which case I will fix up this PR.

xnox:
I think i know what the problem is now.

PR #590, opened 2025-05-23
Remove libpython linkage instead of ignoring it.

### LINKED ISSUE #589: auditwheel is explicitly allowlisting libpython even though it violates the manylinux spec
When auditwheel performs its search for external dependencies, [it currently ignores linkage to libpython](https://github.com/pypa/auditwheel/blob/f9bbbceaf4fa09aaff52432569284a2617092aee/src/auditwheel/policy/__init__.py#L204). This behavior has existed since [the very first commit on the repository](https://github.com/pypa/auditwheel/commit/4562a2928cc6f20d162e102b53b950f8768e91ca#diff-0da1c252b6d01e52c0c332528b155da108c2f6c0520d6fef6f11e885e3bd8a43R15-R82). It also seems like some recent changes like https://github.com/pypa/auditwheel/issues/418 have explicitly chosen to preserve this behavior.

I am not sure why auditwheel makes this choice since [the original manylinux spec in fact explicitly disallows libpython linkage](https://peps.python.org/pep-0513/#libpythonx-y-so-1). That has also been reinforced in more recent iterations of manylinux, e.g. [PEP 599](https://peps.python.org/pep-0599/#the-manylinux2014-policy) says:

> libpythonX.Y remains ineligible for inclusion for the same reasons outlined in PEP 513

Is there a particular reason that auditwheel does not enforce this requirement? Perhaps there are good reasons to exclude this check, at least conditionally.

If not, it should probably be changed, but I would guess that the change might be quite disruptive to projects that have been unknowingly linking to libpython up until now, so such a change would probably need to be rolled out slowly and hidden behind some feature flag to give projects some time to adapt.

I was brought here by https://github.com/astral-sh/python-build-standalone/issues/619. Now that the Python distributions that are installable with uv are statically linking libpython and the DSO is no longer even available in those distributions (precisely the sort of use case that the manylinux requirement is meant to support), various existing wheels that auditwheel would consider valid will fail to load in these environments.

vyasr:
Resolves #589.

I also switched the pre-commit config to use any Python 3 instead of constraining to 3.9. Without that, running pre-commit fails as shown in https://github.com/pre-commit/pre-commit/issues/1375.


src/auditwheel/policy/__init__.py
```diff
@@ -201,9 +200,6 @@ def filter_libs(
                     # 'ld64.so.1' on ppc64le
                     # 'ld-linux*' on other platforms
                     continue
-                if LIBPYTHON_RE.match(lib):
-                    # always exclude libpythonXY
-                    continue
```
mayeut, line 206: This has the effect of checking version symbols in libpython which we don't want to do.

mayeut: This is fixed with the latest commits.
While a test has been added to check that a wheel with libpython is repaired to the correct glibc limit, the test  depends on the environment for failure...
Checked manually inside a `ghcr.io/mayeut/manylinux_2_35:2025.05.19-1` container.

tests/integration/test_bundled_wheels.py
```diff
@@ -170,13 +181,37 @@ def test_wheel_source_date_epoch(timestamp, tmp_path, monkeypatch):
...
+        WHEEL_FILE=[wheel],
+        EXCLUDE=[],
+        DISABLE_ISA_EXT_CHECK=False,
+        ZIP_COMPRESSION_LEVEL=6,
+        cmd="repair",
+        func=Mock(),
+        prog="auditwheel",
+        verbose=0,
+    )
+    main_repair.execute(args, Mock())
```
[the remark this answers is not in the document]
mayeut: patchelf is part of the `test` dependency group. It should always be present to run tests.

src/auditwheel/patcher.py
```diff
@@ -57,6 +60,15 @@ def replace_needed(self, file_name: Path, *old_new_pairs: tuple[str, str]) -> No
             ]
         )
 
+    def remove_needed(self, file_name: Path, *sonames: str) -> None:
```
lkollar, line 63: I agree, this is the right approach.

lkollar (approved):
LGTM. I was wondering if we should introduce some sort of temporary override to disable this behaviour in case it causes issues, but it's always possible to pin to an older version, so I guess that's good enough.

PR #662, opened 2025-12-31
fix: adjust TestPolicyAccess::test_get_by_name for riscv64

cjwatson:
The older `manylinux` policies don't exist on this architecture.


mayeut (changes requested):
Thanks,
Same suggestion for `loongarch64` while this is being update.

Longer run (i.e. not for now), we'll need to check what we want to test with this and either simplify and/or parametrize rather than depend on architecture detection.

PR #672, opened 2026-01-15
Set rpath when a lib has invalid ones

### LINKED ISSUE #671: Do not set rpath to `$ORIGIN` for libs with non-exist rpath
Given a wheel with following libs.

When handling `b.so`, `auditwheel` will only set its rpath when `b.so` has a valid rpath:

https://github.com/pypa/auditwheel/blob/5ef7bd5a17fd7b2487fc6c9b16aca9f23e0d0a5e/src/auditwheel/lddtree.py#L248

https://github.com/pypa/auditwheel/blob/5ef7bd5a17fd7b2487fc6c9b16aca9f23e0d0a5e/src/auditwheel/repair.py#L186-L187

When rpath of `b.so` is not empty, we should also replace it with `$ORIGIN` no matter if it's valid.

```
a.so -> b.so -> c.so
a.so -> c.so

a.so: rpath=$ORIGIN
b.so: runpath=<nonexist>
```

oraluben:
Closes #671

### DIFF
--- a/src/auditwheel/elfutils.py
+++ b/src/auditwheel/elfutils.py
@@ -117,9 +117,13 @@ def elf_read_rpaths(fn: Path) -> dict[str, list[str]]:
 
         for t in section.iter_tags():
             if t.entry.d_tag == "DT_RPATH":
-                result["rpaths"] = parse_ld_paths(t.rpath, root="/", path=str(fn), keep_non_exist=True)
+                result["rpaths"] = parse_ld_paths(
+                    t.rpath, root="/", path=str(fn), keep_non_exist=True
+                )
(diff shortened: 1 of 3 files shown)


mayeut:
Thanks for the PR and sorry for the delay, maintainer time is scarce.
Per https://github.com/pypa/auditwheel/pull/643#discussion_r3069170602, there might be another route to follow, the gist of the pseudo-code would be:

```
if has_other_grafted_deps(dest_path):
    patcher.set_rpath(dest_path, "$ORIGIN")
else:
    patcher.clear_rpath(dest_path)
```


cc @lkollar for further insight/comments on this proposal.

PR #682, opened 2026-03-24
feat(repair): add --strip-level option for granular strip control

### LINKED ISSUE #491: Allow repair to use `strip` with `--strip-debug` instead of only allowing `--strip-all`
Auditwheels allowed the ability to strip symbols on:
* https://github.com/pypa/auditwheel/pull/255

When we use the `--strip` option:
https://github.com/pypa/auditwheel/blob/d6e5b6945d7dc01a6e560bfbd5887ac89377b61f/src/auditwheel/main_repair.py#L81-L87
The option used today is to execute  `strip -s`, see: 
https://github.com/pypa/auditwheel/blob/d6e5b6945d7dc01a6e560bfbd5887ac89377b61f/src/auditwheel/repair.py#L127

We have been doing some debugging on the Apache Arrow project and we would like to use the option to strip only debug symbols in order to preserve the full backtrace including non-public functions when debugging and still be able to reduce the size of our wheels, see more information here: https://github.com/apache/arrow/issues/40749#issuecomment-2045586715

I would like to have the ability to pass the `--strip-debug` to `strip`. I am happy to create a PR if you think this is ok to be done.

taegyunkim:
## Summary

Closes #491. 

- Adds a new `--strip-level` option to `auditwheel repair` with four levels: `none` (default), `debug` (`strip -g`), `unneeded` (`strip --strip-unneeded`), and `all` (`strip -s`)
- Deprecates the existing `--strip` flag (maps to `--strip-level=all` for backward compatibility)
- Validates conflicting `--strip` and `--strip-level` usage early, before any wheel processing begins

### Motivation

The current `--strip` flag only supports `strip -s` (strip all symbols), which is too aggressive for many use cases. Users building wheels with debug symbols want the ability to strip only debug info (`-g`) or unneeded symbols (`--strip-unneeded`) while keeping the symbol table intact for debugging and profiling.

### Design

- `StripLevel` enum in `repair.py` maps each level to the corresponding `strip` CLI arguments
- `process_symbols()` applies stripping to both vendored libraries and extension modules
- Conflict guard and strip normalization happen at the top of `repair_wheel()`, before the pure-wheel early return
- `_get_strip_args(StripLevel.NONE)` raises `ValueError` as a safety net (callers skip processing for `NONE`)

## Test plan

- [x] Unit tests for argument parsing (`TestStripLevelArgument`)
- [x] Unit tests for `execute()` integration (`TestStripLevelExecute`) — verifies strip level is passed through, deprecation warning is emitted, and conflicting args error
- [x] Unit tests for `repair_wheel()` behavior (`TestRepairWheelStripLevels`) — verifies `process_symbols` is called/skipped correctly per level
- [x] Unit tests for `_get_strip_args` and `process_symbols` (`TestGetStripArgs`, `TestProcessSymbols`, `TestStripSymbolsShim`)
- [x] Conflict guard test on pure wheels (`TestRepairWheelConflict`)
- [x] All 200 unit tests pass


src/auditwheel/repair.py
```diff
@@ -37,9 +48,20 @@ def repair_wheel(
     *,
     update_tags: bool,
     patcher: ElfPatcher,
-    strip: bool,
-    zip_compression_level: int,
+    strip: bool | None = None,
+    strip_level: StripLevel | None = None,
+    zip_compression_level: int = zlib.Z_DEFAULT_COMPRESSION,
```
mayeut, line 53: auditwheel is a CLI tool as mentioned in the README. Although the modules do not follow the conventional naming for this, there's no public python API, everything is private. Please do not keep the compatibility shims in the python code.


```suggestion
    strip_level: StripLevel,
    zip_compression_level: int,
```

src/auditwheel/repair.py
```diff
@@ -37,9 +48,20 @@ def repair_wheel(
...
+    # Validate and normalize strip arguments before doing any work.
+    if strip is not None and strip_level is not None:
+        msg = "Cannot specify both 'strip' and 'strip_level' parameters"
+        raise ValueError(msg)
+
+    if strip is True:
+        strip_level = StripLevel.ALL
+    elif strip_level is None:
+        strip_level = StripLevel.NONE
+
```
mayeut, line 64: for the reason above, this is not needed without compatibility in mind.

```suggestion
```

src/auditwheel/repair.py
```diff
@@ -128,10 +150,11 @@ def repair_wheel(
...
             output_wheel = add_platforms(ctx, abis, get_replace_platforms(abis[0]))
 
-        if strip:
-            libs_to_strip = [path for (_, path) in soname_map.values()]
-            extensions = external_refs_by_fn.keys()
-            strip_symbols(itertools.chain(libs_to_strip, extensions))
+        if strip_level != StripLevel.NONE:
+            libs_to_process = [path for (_, path) in soname_map.values()]
+            extensions = list(external_refs_by_fn.keys())
+            all_libraries = list(itertools.chain(libs_to_process, extensions))
```
mayeut, line 156: no need to enforce lists, iterables should be enough
```suggestion
            extensions =external_refs_by_fn.keys()
            all_libraries = itertools.chain(libs_to_process, extensions)
```

src/auditwheel/repair.py
```diff
@@ -128,10 +150,11 @@ def repair_wheel(
...
+            process_symbols(all_libraries, strip_level)
```
mayeut, line 157:
```suggestion
            strip_symbols(all_libraries, strip_level)
```

src/auditwheel/repair.py
```diff
@@ -148,10 +171,41 @@ def repair_wheel(
     return output_wheel
 
 
+def process_symbols(
```
mayeut, line 174:
```suggestion
def strip_symbols(
```

src/auditwheel/repair.py
```diff
@@ -148,10 +171,41 @@ def repair_wheel(
...
+    msg = f"_get_strip_args called with unsupported strip level: {strip_level!r}"
+    raise ValueError(msg)
+
+
 def strip_symbols(libraries: Iterable[Path]) -> None:
-    for lib in libraries:
-        logger.info("Stripping symbols from %s", lib)
-        check_call(["strip", "-s", lib])
+    """Legacy function for backward compatibility. Use process_symbols instead."""
+    process_symbols(libraries, StripLevel.ALL)
```
mayeut, line 208:
```suggestion
```

mayeut (changes requested):
Thanks for your pull request,
Can you please address my inline comments as well as copilot ones ?

PR #683, opened 2026-03-24
Skip internal non-Python references

### LINKED ISSUE #531: Can't repair wheel archive
For google/or-tools we are migrating from `manylinux2014` to `quay.io/pypa/manylinux_2_28_x86_64`

[ortools-9.12.4478-cp312-cp312-linux_x86_64.whl.zip](https://github.com/user-attachments/files/18480462/ortools-9.12.4478-cp312-cp312-linux_x86_64.whl.zip) (ed to remove `.zip` extension needed to upload on github)

General Layout of the wheel package:

```
* ortools:
  * .libs/
    *  libortools.so # our main C++ solver everything in a big fat dynamic library #  RUNPATH $ORIGIN
    * libxxx.so # various third party needed by libortools.so (e.g abseil-cpp, protobuf, etc..)
  * graph/ # module
    * python/ # pybind11 wrapper
      *   /max_flow.cpython-312-x86_64-linux-gnu.so # RUNPATH $ORIGIN:$ORIGIN/../../../ortools/.libs

```
note: take a look at the `ortools/graph/python/max_flow.cpython-312-x86_64-linux-gnu.so`
and `objdump -p ...`

## Protocol

```sh
auditwheel -v repair --plat "manylinux_2_28_x86_64" ortools-9.12.4478-cp312-cp312-linux_x86_64.whl
```

## Expected

A `ortools-9.12.4478-cp312-cp312-manylinux_2_28_x86_64.whl` generated file

## Observed

auditwheel failed...

```sh
...
DEBUG:auditwheel.policy:Package requires GLIBC_2.26, incompatible with policy manylinux_2_24_x86_64 which requires {'GLIBC_2.3.3', 'GLIBC_2.7', 'GLIBC_2.13', 'GLIBC_2.8', 'GLIBC_2.11', 'GLIBC_2.4', 'GLIBC_2.2.6', 'GLIBC_2.24', 'GLIBC_2.3', 'GLIBC_2.15', 'GLIBC_2.9', 'GLIBC_2.2.5', 'GLIBC_2.3.4', 'GLIBC_2.23', 'GLIBC_2.17', 'GLIBC_2.12', 'GLIBC_2.10', 'GLIBC_2.5', 'GLIBC_2.6', 'GLIBC_2.22', 'GLIBC_2.14', 'GLIBC_2.3.2', 'GLIBC_2.16', 'GLIBC_2.18'}
DEBUG:auditwheel.policy:Package requires GLIBC_2.25, incompatible with policy manylinux_2_24_x86_64 which requires {'GLIBC_2.3.3', 'GLIBC_2.7', 'GLIBC_2.13', 'GLIBC_2.8', 'GLIBC_2.11', 'GLIBC_2.4', 'GLIBC_2.2.6', 'GLIBC_2.24', 'GLIBC_2.3', 'GLIBC_2.15', 'GLIBC_2.9', 'GLIBC_2.2.5', 'GLIBC_2.3.4', 'GLIBC_2.23', 'GLIBC_2.17', 'GLIBC_2.12', 'GLIBC_2.10', 'GLIBC_2.5', 'GLIBC_2.6', 'GLIBC_2.22', 'GLIBC_2.14', 'GLIBC_2.3.2', 'GLIBC_2.16', 'GLIBC_2.18'}
DEBUG:auditwheel.policy:Package requires GCC_7.0.0, incompatible with policy manylinux_2_24_x86_64 which requires {'GCC_3.4.2', 'GCC_4.7.0', 'GCC_3.3.1', 'GCC_4.3.0', 'GCC_3.4', 'GCC_4.8.0', 'GCC_3.3', 'GCC_4.0.0', 'GCC_4.2.0', 'GCC_3.4.4', 'GCC_3.0'}
DEBUG:auditwheel.policy:Package requires CXXABI_1.3.11, incompatible with policy manylinux_2_24_x86_64 which requires {'CXXABI_1.3.10', 'CXXABI_1.3.6', 'CXXABI_FLOAT128', 'CXXABI_1.3.8', 'CXXABI_1.3.3', 'CXXABI_1.3.1', 'CXXABI_1.3.4', 'CXXABI_TM_1', 'CXXABI_1.3', 'CXXABI_1.3.9', 'CXXABI_1.3.7', 'CXXABI_1.3.2', 'CXXABI_1.3.5'}
  File "/tmp/py12/bin/auditwheel", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/tmp/py12/lib/python3.12/site-packages/auditwheel/main.py", line 54, in main
    return args.func(args, p)
           ^^^^^^^^^^^^^^^^^^
  File "/tmp/py12/lib/python3.12/site-packages/auditwheel/main_repair.py", line 178, in execute
    out_wheel = repair_wheel(
                ^^^^^^^^^^^^^
  File "/tmp/py12/lib/python3.12/site-packages/auditwheel/repair.py", line 82, in repair_wheel
    raise ValueError(msg)
ValueError: Cannot repair wheel, because required library "libabsl_spinlock_wait.so.2407.0.0" could not be located
```

note: Using the same "workflow" but with `quay.io/pypa/musllinux_1_2_x86_64:latest` works as expected

lkollar:
This should fix #531. While investigating, also discovered a subtle bug when auditwheel can use the wrong manylinux tag when repairing the wheel.

### DIFF
--- a/src/auditwheel/repair.py
+++ b/src/auditwheel/repair.py
@@ -40,7 +40,7 @@ def repair_wheel(
     strip: bool,
     zip_compression_level: int,
 ) -> Path | None:
-    external_refs_by_fn = wheel_abi.full_external_refs
+    external_refs_by_fn = wheel_abi.repair_external_refs
     # Do not repair a pure wheel, i.e. has no external refs
     if not external_refs_by_fn:
         return None
--- a/src/auditwheel/wheel_abi.py
+++ b/src/auditwheel/wheel_abi.py
@@ -206,9 +206,7 @@ def get_wheel_elfdata(
             log.warning("couldn't detect wheel libc, defaulting to %s", str(libc))
             policies = WheelPolicies(libc=libc, arch=architecture)
 
-        wheel_sonames = frozenset(
-            fn.name for fn in itertools.chain(full_elftree, nonpy_elftree)
-        )
+        wheel_sonames = frozenset(fn.name for fn in itertools.chain(full_elftree, nonpy_elftree))
 
         for fn, external_refs in full_external_refs.items():
             filtered_refs = _filter_repair_refs(external_refs, wheel_sonames)
--- a/src/auditwheel/wheel_abi.py
+++ b/src/auditwheel/wheel_abi.py
@@ -35,6 +35,7 @@ log = logging.getLogger(__name__)
 class WheelAbIInfo:
     policies: WheelPolicies
     full_external_refs: dict[Path, dict[str, ExternalReference]]
+    repair_external_refs: dict[Path, dict[str, ExternalReference]]
     overall_policy: Policy
     external_refs: dict[str, ExternalReference]
     ref_policy: Policy
@@ -52,9 +53,32 @@ class WheelElfData:
     policies: WheelPolicies
     full_elftree: dict[Path, DynamicExecutable]
     full_external_refs: dict[Path, dict[str, ExternalReference]]
+    repair_external_refs: dict[Path, dict[str, ExternalReference]]
     versioned_symbols: dict[str, set[str]]
     uses_ucs2_symbols: bool
     uses_pyfpe_jbuf: bool
+
+
+def _filter_repair_refs(
+    external_refs: dict[str, ExternalReference],
+    wheel_sonames: frozenset[str],
+) -> dict[str, ExternalReference]:
+    """Return external refs with unresolved sonames already present in the
+    wheel removed.
+
+    ``external_refs`` maps policy name to the libraries needed by one ELF.
+    ``wheel_sonames`` is the set of shared-library sonames already present in
+    the wheel. The returned mapping keeps only refs that still need repair.
+    """
+    filtered_refs = {}
+    for name, external_ref in external_refs.items():
+        libs = {
+            lib: path
+            for lib, path in external_ref.libs.items()
+            if path is not None or lib not in wheel_sonames
+        }
+        filtered_refs[name] = ExternalReference(libs, external_ref.blacklist, external_ref.policy)
+    return filtered_refs
 
 
 @functools.lru_cache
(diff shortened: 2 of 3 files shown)


mayeut:
> While investigating, also discovered a subtle bug when auditwheel can use the wrong manylinux tag when repairing the wheel.

I fail to understand the issue, I guess this is related to the `target_abi` addition ?
`abis` are always sorted - this should likely be documented and/or enforced - so we do want to repair for `abis[0]` if we're going to add multiple tags as manylinux_x_y1 is always compatible with  manylinux_x_y2 when y1<y2 but the reverse is not true.
Can this be split in a different PR ?

I'm trying to understand what the new filtering is doing exactly. Do you have an example before/after to ease this (or an integration test) ?

lkollar:
You're right, that change is unrelated. I'll move it to a new PR and try to come up with test to reproduce the issue.

lkollar:
@mayeut I filed https://github.com/pypa/auditwheel/issues/690 which hopefully better explains what I think the issue is here. It's quite subtle and I don't think it has a huge impact, but it might be worth fixing it.

mayeut:
@lkollar, if I understand the issue correctly, the problem comes from the fact that `ldd` returns some `None` path in `nonpy_elftree` libraries which would correctly be loaded at runtime because they are loaded as dependencies of a library with a run path that allows to find them (and are found when looking at the elf tree for said library).

Rather than introducing `repair_external_refs` - which seems to assume every soname already in the wheel will be found (if I understand correctly) - I think we might want to "fix-up" the elf tree and only assume it can be found if it's indeed found in any of the elf trees. We can try to come up with one or more alternatives to this PR if you think that makes sense.

mayeut:
@lkollar, I opened an alternative PR for discussion. I checked that it should fix #531.

PR #685, opened 2026-03-27
feat: add --json output flag to `auditwheel show`

lcarva:
Enable scripting by adding a --json flag that outputs structured JSON instead of human-readable text, including wheel tag, compatibility flags, versioned symbols, external libs, and policy upgrade info.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -42,6 +42,7 @@ repos:
       - pytest
       - types-docker
       - types-requests
+      - types-jsonschema
 
 - repo: https://github.com/zizmorcore/zizmor-pre-commit
   rev: e3eebf65325ccc992422292cb7a4baee967cf815  # frozen: v1.26.1
--- a/src/auditwheel/main_show.py
+++ b/src/auditwheel/main_show.py
@@ -9,6 +9,8 @@ from auditwheel import options
 if TYPE_CHECKING:
     import argparse
 
+    from auditwheel.wheel_abi import WheelAbIInfo
+
 logger = logging.getLogger(__name__)
 
 
@@ -16,6 +18,13 @@ def configure_parser(sub_parsers: Any) -> None:  # noqa: ANN401
     help_ = "Audit a wheel for external shared library dependencies."
     p = sub_parsers.add_parser("show", help=help_, description=help_)
     p.add_argument("WHEEL_FILE", type=Path, help="Path to wheel file.")
+    p.add_argument(
+        "--json",
+        dest="JSON",
+        action="store_true",
+        default=False,
+        help="Output results in JSON format",
+    )
     options.disable_isa_check(p)
     options.allow_pure_python_wheel(p)
     options.ldpaths(p)
@@ -27,6 +36,42 @@ def printp(text: str) -> None:
 
     print()
     print("\n".join(wrap(text, break_long_words=False, break_on_hyphens=False)))
+
+
+def _output_json(fn: str, winfo: WheelAbIInfo) -> None:
+    from auditwheel import json
+
+    policies = winfo.policies
+    libs = winfo.external_refs[policies.lowest.name].libs
+
+    policy_upgrades: dict[str, dict[str, Any]] = {}
+    for p in policies:
+        if p > winfo.overall_policy:
+            entry: dict[str, Any] = {}
+            p_libs = winfo.external_refs[p.name].libs
+            if len(p_libs):
+                entry["libs_to_eliminate"] = sorted(p_libs.keys())
+            blacklist = winfo.external_refs[p.name].blacklist
+            if len(blacklist):
+                entry["blacklisted_symbols"] = {k: sorted(v) for k, v in sorted(blacklist.items())}
+            if entry:
+                policy_upgrades[p.name] = entry
+
+    result: dict[str, Any] = {
+        "version": 1,
+        "wheel": fn,
+        "pure": False,
+        "overall_tag": winfo.overall_policy.name,
+        "sym_tag": winfo.sym_policy.name,
+        "pyfpe": winfo.pyfpe_policy == policies.linux,
+        "ucs2": winfo.ucs_policy == policies.linux,
+        "unsupported_isa": winfo.machine_policy == policies.linux,
+        "versioned_symbols": {k: sorted(v) for k, v in sorted(winfo.versioned_symbols.items())},
+        "external_libs": {str(k): str(v) if v else None for k, v in sorted(libs.items())},
+        "policy_upgrades": policy_upgrades,
+    }
+
+    print(json.dumps(result))
 
 
 def execute(args: argparse.Namespace, parser: argparse.ArgumentParser) -> int:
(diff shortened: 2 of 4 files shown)


mayeut (changes requested):
Not reviewed thoroughly yet but left a couple comments inlined in addition to the following more general comment.

Per https://github.com/pypa/auditwheel/issues/676#issuecomment-4015889723, as the json output should be an "API", it should be versioned, and, ideally, we might want to have a json schema for this.

src/auditwheel/main_show.py
```diff
@@ -76,10 +120,19 @@ def execute(args: argparse.Namespace, parser: argparse.ArgumentParser) -> int:
         )
     except NonPlatformWheelError as e:
         logger.info("%s", e.message)
+        if args.JSON:
```
lcarva: I'm not sure if handling this special json case is important. Also, the case below where pure wheels are allowed does not emit a json. What's the preference?

src/auditwheel/main_show.py
```diff
@@ -38,6 +47,41 @@ def printp(text: str) -> None:
     print("\n".join(wrap(text, break_long_words=False, break_on_hyphens=False)))
 
 
+def _output_json(fn: str, winfo: WheelAbIInfo) -> int:
```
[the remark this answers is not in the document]
lcarva: Done!

.pre-commit-config.yaml
```diff
@@ -33,7 +33,7 @@ repos:
   rev: v1.19.1
   hooks:
   - id: mypy
-    exclude: tests/integration/.*/.*|tests/integration/quick_check_numpy.py|tests/unit/test_wheel_abi.py|scripts/calculate_symbol_versions.py
+    exclude: tests/integration/.*/.*|tests/integration/quick_check_numpy.py|tests/unit/test_wheel_abi.py|tests/unit/test_main_show.py|scripts/calculate_symbol_versions.py
```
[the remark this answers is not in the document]
lcarva: Understood! I was getting a linting error after adding `types-jsonschema`. The fix is in a separate commit here but also available on its own here: https://github.com/pypa/auditwheel/pull/687

src/auditwheel/main_show.py
mayeut: > What's the preference?

If adding this, let's try to add it properly.
If pure wheels are allowed then we should not error but output a specific json output for pure wheels, e.g. the schema could have something like:
```
  "oneOf": [
    { "$ref": "#/$defs/result_pure" },
    { "$ref": "#/$defs/result_platform" },
    { "$ref": "#/$defs/error" }
  ],
```
adding a "pure" boolean property that's always true for result_pure and always false for result_platform

.pre-commit-config.yaml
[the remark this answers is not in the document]
mayeut: #687 merged, thanks

src/auditwheel/main_show.py
lcarva: Yeah, that's a good idea! Implemented that approach. (I also removed the unnecessary .. import json which already happens earlier in this function.)

lcarva:
@mayeut, please let me know if there's something I can do to help move this work forward (assuming that there is an interest in doing so.)

mayeut:
@lcarva, there's still interest however I'd like to get a couple fixes & the android support so it might take a bit of time before this can be reviewed again.

tonisbones:
Would love to see this PR get merged :)

mayeut:
@lcarva, can the copilot reviews be answered/addressed please ?

lcarva:
@mayeut, I believe those are all addressed now! Sorry for the delay on this.

src/auditwheel/main_show.py
```diff
@@ -29,6 +38,42 @@ def printp(text: str) -> None:
     print("\n".join(wrap(text, break_long_words=False, break_on_hyphens=False)))
 
 
+def _output_json(fn: str, winfo: WheelAbIInfo) -> None:
+    from auditwheel import json
+
+    policies = winfo.policies
+    libs = winfo.external_refs[policies.lowest.name].libs
```
mayeut: I missed that one before.
Is it really useful given what's considered external libs is dependant on the policy and that should be covered by policy_upgrades below ?

lcarva: external_libs gives you the path of the system libraries which is what I'm mostly interested in :smile: 
policy_upgrades just gives you a list of libraries, e.g.:
```
{
    "version": 1,
    "wheel": "testdependencies-0.0.1-cp313-cp313-linux_x86_64.whl",
    "pure": false,
    "overall_tag": "linux_x86_64",
    "sym_tag": "manylinux_2_39_x86_64",
    "pyfpe": false,
    "ucs2": false,
    "unsupported_isa": false,
    "versioned_symbols": {
        "libc.so.6": [
            "GLIBC_2.2.5",
            "GLIBC_2.39"
        ]
    },
    "lowest_policy": "manylinux_2_41_x86_64",
    "external_libs": {
        "libdependency.so": "/home/lucarval/src/github.com/pypa/auditwheel/tests/integration/testdependencies/libdependency.so"
    },
    "policy_upgrades": {
        "manylinux_2_41_x86_64": {
            "libs_to_eliminate": [
                "libdependency.so"
            ]
        },
        "manylinux_2_40_x86_64": {
            "libs_to_eliminate": [
                "libdependency.so"
            ]
        },
        "manylinux_2_39_x86_64": {
            "libs_to_eliminate": [
                "libdependency.so"
            ]
        },
        "manylinux_2_38_x86_64": {
            "libs_to_eliminate": [
                "libdependency.so"
            ]
        },
        "manylinux_2_37_x86_64": {
            "libs_to_eliminate": [
                "libdependency.so"
            ]
        },
        "manylinux_2_36_x86_64": {
            "libs_to_eliminate": [
                "libdependency.so"
            ]
        },
        "manylinux_2_35_x86_64": {
            "libs_to_eliminate": [
                "libdependency.so"
            ]
        },
        "manylinux_2_34_x86_64": {
            "libs_to_eliminate": [
                "libdependency.so"
            ]
        },
        "manylinux_2_31_x86_64": {
            "libs_to_eliminate": [
                "libdependency.so"
            ]
        },
        "manylinux_2_28_x86_64": {
            "libs_to_eliminate": [
                "libdependency.so"
            ]
        },
        "manylinux_2_27_x86_64": {
            "libs_to_eliminate": [
                "libdependency.so"
            ]
        },
        "manylinux_2_26_x86_64": {
            "libs_to_eliminate": [
                "libdependency.so"
            ]
        },
        "manylinux_2_24_x86_64": {
            "libs_to_eliminate": [
                "libdependency.so"
            ]
        },
        "manylinux_2_17_x86_64": {
            "libs_to_eliminate": [
                "libdependency.so"
            ]
        },
        "manylinux_2_12_x86_64": {
            "libs_to_eliminate": [
                "libdependency.so"
            ]
        },
        "manylinux_2_5_x86_64": {
            "libs_to_eliminate": [
                "libdependency.so"
            ]
        }
    }
}
```

Picking from `policies.lowest.name` mirrors the data in the non json output. I added a commit to this PR that clarifies what the lowest policy is (`"lowest_policy": "manylinux_2_41_x86_64",`). That should help disambiguate some of this.

How does that sound?

mayeut: > external_libs gives you the path of the system libraries which is what I'm mostly interested in

In this case I'd go for highest policy which includes all possible libraries (lowest policy might not include all of them, e.g. `libmvec.so.1`) and remove the added `lowest_policy`.

lcarva: Done!

(BTW, lint is broken on main, https://github.com/pypa/auditwheel/pull/713 should fix it.)

mayeut: thanks for the fix

mayeut (approved):

lkollar (approved):

mayeut:
Thanks @lcarva

PR #692, opened 2026-04-18
fix: non python ELF tree dependencies resolution

### LINKED ISSUE #531: Can't repair wheel archive

mayeut:
auditwheel should assumes that if a dependency in a non python ELF tree can be resolved in any other fully resolved  ELF tree then, it can be loaded from other non python ELF trees as well.

Python extensions ELF trees should be not fixed up, they should always be valid in the first place.

The rationale for this difference between Python extensions ELF trees and non python ELF trees is that we know that python extensions ELF are "top-level" and should be able to load all their dependencies. We can't assume non python ELF files are "top-level" and thus fix-up resolution using other ELF trees if we can.

Missing tests for now

close #683
fix #531

### DIFF
--- a/src/auditwheel/wheel_abi.py
+++ b/src/auditwheel/wheel_abi.py
@@ -58,6 +58,58 @@ class WheelElfData:
     uses_pyfpe_jbuf: bool
 
 
+def _fixup_elf_trees(
+    py_elf_trees: dict[Path, DynamicExecutable],
+    nonpy_elftree: dict[Path, DynamicExecutable],
+) -> None:
+    # Auditwheel assumes that if a dependency in a non-Python ELF tree can be resolved
+    # in any other fully resolved ELF tree, then it can be loaded from other
+    # non-Python ELF trees as well. Python extension ELF trees are not
+    # fixed up; they should always be valid in the first place. We fix up
+    # nonpy_elftree with that assumption in mind.
+    resolved_executables = [
+        executable
+        for executable in itertools.chain(py_elf_trees.values(), nonpy_elftree.values())
+        if all(library.realpath is not None for library in executable.libraries.values())
+    ]
+
+    def _get_reference(path_to_elf: Path) -> DynamicExecutable | None:
+        for candidate in resolved_executables:
+            for library in candidate.libraries.values():
+                assert library.realpath is not None  # noqa: S101
+                if path_to_elf.samefile(library.realpath):
+                    return candidate
+        return None
+
+    for elf_path, elf_executable in nonpy_elftree.items():
+        need_resolution = any(
+            elf_library.realpath is None for elf_library in elf_executable.libraries.values()
+        )
+        if not need_resolution:
+            continue
+        reference = _get_reference(elf_executable.realpath)
+        if reference is None:
+            continue
+        libraries_new: dict[str, DynamicLibrary] = {}
+        soname_all = set(elf_executable.libraries.keys())
+        for soname, elf_library in elf_executable.libraries.items():
+            if elf_library.realpath is None and soname in reference.libraries:
+                libraries_new[soname] = reference.libraries[soname]
+                # we need to add newly found dependencies as well
+                needed_new = set(libraries_new[soname].needed) - soname_all
+                while needed_new:
+                    needed_new_copy = needed_new.copy()
+                    for soname_needed in needed_new_copy:
+                        if soname_needed in reference.libraries:
+                            libraries_new[soname_needed] = reference.libraries[soname_needed]
+                            soname_all.add(soname_needed)
+                            needed_new.update(libraries_new[soname_needed].needed)
+                    needed_new -= soname_all
+            else:
+                libraries_new[soname] = elf_library
+        nonpy_elftree[elf_path] = dataclasses.replace(elf_executable, libraries=libraries_new)
+
+
 @functools.lru_cache
 def get_wheel_elfdata(
     libc: Libc | None,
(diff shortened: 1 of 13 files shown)


src/auditwheel/wheel_abi.py
```diff
@@ -164,6 +165,52 @@ def get_wheel_elfdata(
             arch = None if architecture is None else architecture.value
             raise NonPlatformWheelError(arch, shared_libraries_with_invalid_machine)
 
+        # Auditwheel assumes that if a dependency in a non-Python ELF tree can
+        # be resolved in any other ELF tree, then it can be loaded from other
+        # non-Python ELF trees as well. Python extension ELF trees are not
+        # fixed up; they should always be valid in the first place. We fix up
+        # nonpy_elftree with that assumption in mind.
```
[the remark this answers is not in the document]
mayeut: will likely do that when adding tests if that's the direction we choose to follow

src/auditwheel/wheel_abi.py
```diff
@@ -164,6 +165,52 @@ def get_wheel_elfdata(
...
+                        needed_new -= soname_all
+                    changed = True
+                else:
+                    libraries_new[soname] = elf_library
+            if changed:
+                nonpy_elftree[elf_path] = dataclasses.replace(
+                    elf_executable,
+                    libraries=libraries_new,
+                )
+
```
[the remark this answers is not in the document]
mayeut: will add tests if that's the direction we choose to follow

src/auditwheel/wheel_abi.py
```diff
@@ -164,6 +165,52 @@ def get_wheel_elfdata(
...
+            for soname, elf_library in elf_executable.libraries.items():
+                if elf_library.realpath is None and soname_library_map[soname].realpath is not None:
+                    libraries_new[soname] = soname_library_map[soname]
+                    # we need to add newly found dependencies as well
+                    needed_new = set(libraries_new[soname].needed) - soname_all
+                    while needed_new:
+                        needed_new_copy = needed_new.copy()
+                        for soname_needed in needed_new_copy:
+                            if soname_needed not in soname_all:
+                                libraries_new[soname_needed] = soname_library_map[soname_needed]
```
lkollar, line 201: There is no check to see if `soname_needed` is in the dict so this could lead to a crash. I think this could happen if a library is excluded in [lddtree.py](https://github.com/pypa/auditwheel/blob/e296dd6164f9d16552dd46248f38a45e8d279818/src/auditwheel/lddtree.py#L574) and ends up in `needed` but not in `sonema_library_map`.

src/auditwheel/wheel_abi.py
```diff
@@ -164,6 +165,52 @@ def get_wheel_elfdata(
...
+        # fixed up; they should always be valid in the first place. We fix up
+        # nonpy_elftree with that assumption in mind.
+        soname_library_map: dict[str, DynamicLibrary] = {}
+        for elf_executable in itertools.chain(full_elftree.values(), nonpy_elftree.values()):
+            for soname, elf_library in elf_executable.libraries.items():
+                if soname in soname_library_map:
+                    realpath = soname_library_map[soname].realpath
+                    if realpath is None:
+                        soname_library_map[soname] = elf_library
+                    elif elf_library.realpath is not None and realpath != elf_library.realpath:
```
lkollar, line 180: If the same SONAME resolves to different paths across ELF trees this will fail. I think previously this was not a hard failure.

lkollar (commented):
I agree, this is a better fix, but it can be riskier as it's a bigger change. We'll need integration tests to make sure we don't cause regressions.

src/auditwheel/wheel_abi.py
mayeut: This should never happen since excluded libraries are filtered out from `needed` but let's add a check and a warning. It's likely we won't be able to cover this. If we can, we have another bug.

src/auditwheel/wheel_abi.py
mayeut: Agreed, even if this is an edge case I hope we wouldn't run into, it's still possible.
That means the approach taken here of building a single map for multiple ELF trees is not the right one.
Let's remove this caching phase and try to find a single ELF tree from which to get resolution information in what is currently the 2nd pass.
Removing this caching phase will also remove all possible regressions for wheels that can already be repaired without errors which is a good thing.

src/auditwheel/wheel_abi.py
mayeut: It can be covered by unit tests but it shouldn't be reachable by integration tests.

src/auditwheel/wheel_abi.py
mayeut: done

src/auditwheel/wheel_abi.py
```diff
@@ -57,6 +58,65 @@ class WheelElfData:
...
+
+    def _get_reference(path_to_elf: Path) -> DynamicExecutable | None:
+        for candidate in resolved_executables:
+            for library in candidate.libraries.values():
+                assert library.realpath is not None  # noqa: S101
+                assert library.realpath.is_file()  # noqa: S101
+                assert path_to_elf.is_file()  # noqa: S101
+                if path_to_elf.samefile(library.realpath):
+                    return candidate
+        return None
```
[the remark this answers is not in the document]
mayeut: building the one time index will add a penalty for all wheels that do not require this feature which is likely the majority.
let's keep optimisation for a follow-up if needed.

src/auditwheel/wheel_abi.py
```diff
@@ -57,6 +58,65 @@ class WheelElfData:
...
+        for executable in itertools.chain(py_elf_trees.values(), nonpy_elftree.values())
+        if all(library.realpath is not None for library in executable.libraries.values())
+    ]
+
+    def _get_reference(path_to_elf: Path) -> DynamicExecutable | None:
+        for candidate in resolved_executables:
+            for library in candidate.libraries.values():
+                assert library.realpath is not None  # noqa: S101
+                assert library.realpath.is_file()  # noqa: S101
+                assert path_to_elf.is_file()  # noqa: S101
```
lkollar: This should be done before the loop.

lkollar (approved):
Just a minor comment, LGTM otherwise.

PR #693, opened 2026-04-18
fix: always set RPATH or clear RUNPATH/RPATH for grafted libraries

### LINKED ISSUE #344: Always set rpath to point at the .libs directory
When grafting a library, it's dependencies are either:
- available on the system (as specified in the policy)
- also grafted into the wheel.

So, we should always set the rpath of the grafted library to point at the .libs directory.

Currently (see https://github.com/pypa/auditwheel/blob/9b0de8d2563dd141132e119e46256ccef2acf48f/src/auditwheel/repair.py#L155):
- we only set the rpath if the library already has rpath or runpath set
- set the rpath to a relative path. Note that relative paths are interpreted by the dynamic linker as relative to the current working directory, not relative to the library. To use a path relative to the library, we need `$ORIGIN`.

So, the patch I'm proposing is:
```
--- repair.py	2021-10-23 04:36:19.144441648 +0000
+++ repair.py	2021-10-23 04:36:38.860464792 +0000
@@ -151,8 +151,7 @@
 
     patcher.set_soname(dest_path, new_soname)
 
-    if any(itertools.chain(rpaths["rpaths"], rpaths["runpaths"])):
-        patcher.set_rpath(dest_path, dest_dir)
+    patcher.set_rpath(dest_path, "$ORIGIN")
 
     return new_soname, dest_path
 ```

### LINKED ISSUE #451: missing RPATH in secondary SO packed into the wheel

### LINKED ISSUE #459: Improvement: just set rpath on repair everytime
Please let me know if you all appreciate these code improvements as I scan the code.

I am reading the code for `repair.py` and I believe it would be simpler to always just squash the `RUNPATH`/`RPATH` [1]

I see a few other improvements that can be made. Happy to upstream some if you find them useful.

[1] https://github.com/pypa/auditwheel/blob/3636dad88d6168aa745684db9ea6a7c6df4d4484/src/auditwheel/repair.py#L165

### LINKED ISSUE #617: RPATHs of grafted libraries should ALWAYS be overridden to $ORIGIN
I just ran into a weird edge case. I have a library which was built on another machine, and has a single entry for its rpath: `/path/that/only/exists/on/other/machine/`. I'm using this on my machine and I can get this library to function because I have the LD_LIBRARY_PATH environment variable set to the proper paths.

What I noticed though is that when this library gets grafted into the wheel, we run through this codepath:
```
    if any(itertools.chain(rpaths["rpaths"], rpaths["runpaths"])):
        patcher.set_rpath(dest_path, "$ORIGIN")

```
The rpaths dictionary is setup so that it parses out any RPATH entries that don't exist on the current machine, so at the time this code is executed, the if statement is false the RPATH is never forced to $ORIGIN.

I think this logic should change to one of the following:
1) Always set rpath to $ORIGIN; get rid of the if check here. If a library has no dependencies, then this is safe because no dependencies are loaded. If the library does have dependencies, then it should be getting it from the current flattened directory anyways. I think this is what we should do.

2) If we don't want to do that, then we should check if it has anything in the rpath or runpath entries, without checking if that directory exists first.

I think getting rid of the if statement is the easiest way to fix this edge case

### LINKED ISSUE #671: Do not set rpath to `$ORIGIN` for libs with non-exist rpath

### PR DESCRIPTION
**Pull request by mayeut**
follow-up on https://github.com/pypa/auditwheel/pull/643#discussion_r3069170602

The [rationale](https://github.com/pypa/auditwheel/issues/451#issuecomment-1722204550) for not setting `RPATH` unconditionally is that every patchelf modification comes with a risk and that it's mostly not needed for runtime.
There are a number of issues requesting that it'd be set anyway.

The proposed solution here is to set `RPATH` when grafted dependency has itself other grafted dependencies and to clear `RUNPATH`/`RPATH` when it's not the case. The 2nd operation should mostly be no-op patchelf-wise thus reducing the risk at least for "leaf" dependencies.

fix #671
fix #617
fix #459
fix #451
fix #344

close #672

### CODE REVIEW

File: src/auditwheel/repair.py [hunk 1]
```diff
@@ -112,14 +112,18 @@ def repair_wheel(
...
         for _, path in soname_map.values():
-            needed = elf_read_dt_needed(path)
+            needed = elf_read_dt_needed(path)  # TODO perf, we already read those at some point
             replacements = []
             for n in needed:
                 if n in soname_map:
                     replacements.append((n, soname_map[n][0]))
             if replacements:
+                patcher.set_rpath(path, "$ORIGIN")
                 patcher.replace_needed(path, *replacements)
```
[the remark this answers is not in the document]
mayeut: this would change the behaviour w.r.t `DT_NEEDED` updates. might be for a follow up.

oraluben (approved):
LGTM

src/auditwheel/repair.py
```diff
@@ -112,14 +112,18 @@ def repair_wheel(
...
+            else:
+                patcher.clear_rpath(path)
```
lkollar, line 126: Can it happen that we clear the RPATH for entries which are internal to the wheel? I don't think those will show up in `soname_map` and this can cause a regression where we clear a RUNPATH which is required to pick up .so's internal to the wheel.

src/auditwheel/repair.py
```diff
@@ -112,14 +112,18 @@ def repair_wheel(
         # they may have internal dependencies (DT_NEEDED) on one another, so
         # we need to update those records so each now knows about the new
         # name of the other.
+        # we also clear or set RPATH depending on the presence of internal dependencies
         for _, path in soname_map.values():
```
lkollar, line 116: I think there can be a logic issue here if the same library is stored under multiple SONAME entries in `soname_map`: 

 ```python
   soname_map["libfoo.so"] = ("libfoo-abc.so", libfoo-abc.so)
   soname_map["libfoo.so.1"] = ("libfoo-abc.so", libfoo-abc.so)
 ```

The first loop iteration will call `set_rpath(path, "$ORIGIN")` and rewrite internal `DT_NEEDED` entries. Processing the second SONAME entry pointing to the same library  `elf_read_dt_needed` will see the already-rewritten name and find no matching replacements, and invoke `clear_rpath(path)`.

One way to solve this would be to perform this in two passes: first record what needs changes and in a second pass, perform the actual changes to RPATH.

src/auditwheel/repair.py
mayeut: This only applies to external grafted libraries so we can't be modifying for entries which are internal to the wheel.

src/auditwheel/repair.py
mayeut: Any given path can only have a single soname so this can't happen.

lkollar (approved):

auvipy (approved):

PR #700, opened 2026-05-30
chore: fix type checking issues

mayeut:
pyelftools recently [released](https://github.com/eliben/pyelftools/issues/660) a version with typing hints.
auditwheel fails to pass mypy checks with this new version.

This PR aims to fix (or ignore) mypy reported issues.

### DIFF
--- a/src/auditwheel/elfutils.py
+++ b/src/auditwheel/elfutils.py
@@ -1,24 +1,41 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, TypeVar
 
 from elftools.common.exceptions import ELFError
+from elftools.elf.dynamic import DynamicSection
 from elftools.elf.elffile import ELFFile
+from elftools.elf.gnuversions import GNUVerNeedSection
+from elftools.elf.sections import Section, SymbolTableSection
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Iterator
     from pathlib import Path
+
+    _SectionT = TypeVar("_SectionT", bound=Section)
+
+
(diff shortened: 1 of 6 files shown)


src/auditwheel/elfutils.py
```diff
@@ -1,24 +1,41 @@
...
+
+def _get_section_by_type(elf: ELFFile, type_: type[_SectionT]) -> _SectionT | None:
+    name = {
+        DynamicSection: ".dynamic",
+        GNUVerNeedSection: ".gnu.version_r",
+        SymbolTableSection: ".dynsym",
+    }[type_]
+    section = elf.get_section_by_name(name)
+    if TYPE_CHECKING:
+        assert section is None or isinstance(section, type_)
```
mayeut, line 26: This could be made a runtime check but this way, we ensure the exact same runtime behaviour as before.

src/auditwheel/elfutils.py
```diff
@@ -1,24 +1,41 @@
...
     needed: list[str] = []
     with fn.open("rb") as f:
         elf = ELFFile(f)
-        section = elf.get_section_by_name(".dynamic")
+        section = _get_section_by_type(elf, DynamicSection)
         if section is None:
             msg = f"Could not find soname in {fn}"
             raise ValueError(msg)
-        needed.extend(t.needed for t in section.iter_tags() if t.entry.d_tag == "DT_NEEDED")
+        needed.extend(t.needed for t in section.iter_tags() if t.entry.d_tag == "DT_NEEDED")  # type: ignore[attr-defined]
```
mayeut, line 38: All the `# type: ignore[attr-defined]` are for dynamically defined attributes

PR #704, opened 2026-06-28
feat: reduce calls to patchelf

mayeut:
patchelf is called multiple times to update soname, remove/replace needed, update RPATH.

This can be done in a single command line call allowing for faster patching and reducing the chance of running into a patchelf issue with multiple updates to a single ELF file.

### DIFF
--- a/src/auditwheel/patcher.py
+++ b/src/auditwheel/patcher.py
@@ -110,6 +110,7 @@ def _verify_patchelf() -> Path:
     if m and tuple(int(x) for x in m.group(1).split(".")) >= (0, 14):
         return Path(patchelf_path)
     msg = f"{version.strip()} found. auditwheel repair requires patchelf >= 0.14."
+    raise ValueError(msg)
 
 
 class Patchelf(ElfPatcher):
--- a/src/auditwheel/patcher.py
+++ b/src/auditwheel/patcher.py
@@ -124,7 +124,10 @@ class Patchelf(ElfPatcher):
         return output.removesuffix(" (legacy)")
 
     def apply_updates(self) -> None:
-        for filepath, update_info in self._updates.items():
+        elf_files = list(self._updates.keys())
+        for filepath in elf_files:
+            # prevent re-applying by removing from self._updates
+            update_info = self._updates.pop(filepath)
             args = []
             if update_info.soname is not None:
                 args.extend(["--set-soname", update_info.soname])
@@ -150,3 +153,4 @@ class Patchelf(ElfPatcher):
                 args.extend(set_args)
             if args:
                 check_call([self._patchelf_path, *args, filepath])
+        assert len(self._updates) == 0  # noqa: S101
--- a/src/auditwheel/patcher.py
+++ b/src/auditwheel/patcher.py
@@ -67,11 +67,11 @@ class ElfPatcher:
 
     def get_rpath(self, file_name: Path) -> str:
         key = str(file_name.resolve(strict=True))
-        update_info = self._updates[key]
-        if update_info.rpath is not None:
-            return update_info.rpath
-        if update_info.clear_rpath:
-            return ""
+        if update_info := self._updates.get(key, None):
+            if update_info.rpath is not None:
+                return update_info.rpath
+            if update_info.clear_rpath:
+                return ""
         return self.get_rpath_direct(file_name)
 
     def clear_rpath(self, file_name: Path) -> None:
--- a/src/auditwheel/patcher.py
+++ b/src/auditwheel/patcher.py
@@ -27,46 +27,42 @@ class ElfUpdateInfo:
 class ElfPatcher:
     def __init__(self, platform: str) -> None:
         self.platform = platform
-        self._updates: dict[str, ElfUpdateInfo] = defaultdict(ElfUpdateInfo)
+        self._updates: dict[Path, ElfUpdateInfo] = defaultdict(ElfUpdateInfo)
 
     @staticmethod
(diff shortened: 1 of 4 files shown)


src/auditwheel/patcher.py
```diff
@@ -1,107 +1,156 @@
...
+        args = ("--print-rpath", file_name)
+        output = check_output([self._patchelf_path, *args]).decode("utf-8").strip()
+        return output.removesuffix(" (legacy)")
+
+    def apply_updates(self) -> None:
+        elf_files = list(self._updates.keys())
+        for filepath in elf_files:
+            # prevent re-applying by removing from self._updates
+            update_info = self._updates.pop(filepath)
+            args = []
```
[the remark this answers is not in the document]
mayeut: no retries possibles, the command line will appear in the exception for debugging/reporting.

mayeut:
This PR comes from the investigation regarding the use lief-patchelf in #664 
There are also bugs in LIEF which are triggered when iteratively patching the same ELF file (not yet reported upstream).

src/auditwheel/patcher.py
```diff
@@ -1,107 +1,156 @@
...
+    rpath: str | None = None
+    remove_needed: list[str] = dataclasses.field(default_factory=list)
+    replace_needed: list[tuple[str, str]] = dataclasses.field(default_factory=list)
+    clear_rpath: bool = False
 
 
 class ElfPatcher:
     def __init__(self, platform: str) -> None:
         self.platform = platform
+        self._updates: dict[str, ElfUpdateInfo] = defaultdict(ElfUpdateInfo)
```
lkollar: Why not store `Path` as the key? In most calls we are already working with a `Path` anyway and it stores more information.
```suggestion
        self._updates: dict[Path, ElfUpdateInfo] = defaultdict(ElfUpdateInfo)
```

src/auditwheel/patcher.py
```diff
@@ -1,107 +1,156 @@
...
+        for soname in removed:
+            for org_soname, new_soname in replaced:
+                if soname in (org_soname, new_soname):
+                    return soname
+        return None
 
     def replace_needed(self, file_name: Path, *old_new_pairs: tuple[str, str]) -> None:
-        raise NotImplementedError
+        key = str(file_name.resolve(strict=True))
+        update_info = self._updates[key]
```
lkollar: This is duplicated a bunch of times. How about adding a small helper to simplify it and reduce duplication (this assumes a `Path`):

```python
def _update_for(self, file_name: Path) -> ElfUpdateInfo:
    return self._updates[file_name.resolve(strict=True)]
```
And then
```python
     update_info = self._update_for(file_name)
```

src/auditwheel/patcher.py
```diff
@@ -1,107 +1,156 @@
...
-    def set_rpath(self, file_name: Path, rpath: str) -> None:
-        self.check_rpath_support()
-
-        set_args: list[str | Path] = ["patchelf", "--force-rpath", "--set-rpath", rpath, file_name]
-        if self.platform.startswith("android"):
-            # Android supports only RUNPATH, not RPATH.
-            set_args.remove("--force-rpath")
-
-        # we only want an RPATH, remove RUNPATH/RPATH altogether in a 1st pass
-        check_call(["patchelf", "--remove-rpath", file_name])
```
lkollar, line 100: I think we are not enforcing this anymore: we only strip RPATH if `clear_rpath` is explicitly called.

src/auditwheel/patcher.py
```diff
@@ -1,107 +1,156 @@
...
 
 
 class ElfPatcher:
     def __init__(self, platform: str) -> None:
         self.platform = platform
+        self._updates: dict[str, ElfUpdateInfo] = defaultdict(ElfUpdateInfo)
+
+    @staticmethod
+    def _needed_overlap(removed: Iterable[str], replaced: Iterable[tuple[str, str]]) -> str | None:
+        for soname in removed:
```
lkollar: A bit cleaner:

```python
replaced_sonames = set(chain.from_iterable(replaced))
return next((name for name in removed if name in replaced_sonames), None)
```

src/auditwheel/patcher.py
```diff
@@ -1,107 +1,156 @@
...
-        return check_output(["patchelf", "--print-rpath", file_name]).decode("utf-8").strip()
-
-    def clear_rpath(self, file_name: Path) -> None:
-        check_call(["patchelf", "--remove-rpath", file_name])
+        self._patchelf_path = str(_verify_patchelf())
+
+    def get_rpath_direct(self, file_name: Path) -> str:
+        args = ("--print-rpath", file_name)
+        output = check_output([self._patchelf_path, *args]).decode("utf-8").strip()
+        return output.removesuffix(" (legacy)")
```
lkollar: I find this a bit more readable:
```suggestion
    output = check_output([self._patchelf_path, "--print-rpath", file_name])
    return output.decode("utf-8").strip().removesuffix(" (legacy)")
```

src/auditwheel/patcher.py
mayeut: I need to check why this did not trigger a test failure.

mayeut: This was a workaround (#173) for a bug in patchelf which was fixed in [patchelf 0.11](https://github.com/NixOS/patchelf/commit/67091ca1f979829df9cdcd3609f72d7e78d8a96a).
Given we now require patchelf>=0.14, I think it's safe to drop this.

lkollar (approved):

PR #708, opened 2026-07-04
test with patchelf==0.19.0.0rc1

mayeut:
just to check with latest manylinux images vs #664 which uses some tags.


mayeut:
xref https://github.com/pypa/manylinux/issues/1966
xref https://discuss.python.org/t/need-help-testing-patchelf-0-19-0-in-manylinux-images/108025

PR #709, opened 2026-07-05
feat: add `--allowed-patchers` option to `auditwheel repair`

mayeut:
This allows to define an ordered list of ELF patchers to try.

towards https://github.com/pypa/auditwheel/issues/174

I'm really not sure if this should be an ordered list or just a specific patcher to use. I was thinking the former at first but I'm leaning towards the latter now.


mayeut:
closing in favour of #711

PR #711, opened 2026-07-11
feat: add `--patcher` option to `auditwheel repair`

mayeut:
Alternative to #709 that does not use a list of allowed patchers but instead requires to specify only one (defaults to `patchelf` ensuring reproducibility).

The list of ELF patchers supported is:
- `patchelf` (default)
- `lief-patchelf`
- `none` (will error out if patching is needed)

Towards #174


tests/unit/test_elfpatcher.py
```diff
@@ -71,19 +72,76 @@ def test_elfpatcher_update_elf_path_noop(tmp_path):
...
+        ("lief-patchelf", _Patchelf),
+    ],
+)
+def test_get_patcher(patcher_name, type_, monkeypatch):
+    monkeypatch.setattr(auditwheel.patcher, "which", lambda x: x)
+    patcher = ElfPatcher.get_patcher(patcher_name)
+    assert type(patcher) is type_
+    if isinstance(patcher, _Patchelf):
+        assert patcher._patchelf_path == patcher_name
+
```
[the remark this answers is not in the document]
mayeut: patchelf is a test dependency.

src/auditwheel/main_repair.py
```diff
@@ -122,6 +122,16 @@ def configure_parser(sub_parsers: Any) -> None:  # noqa: ANN401
     options.disable_isa_check(parser)
     options.allow_pure_python_wheel(parser)
     options.ldpaths(parser)
+    parser.add_argument(
```
lkollar, line 125: Should we validate that the selected patcher is valid for the given platform? The tests already do this. We could set the compatible platforms on the `Patcher` class to define these with the patcher implementation.

src/auditwheel/patcher.py
```diff
@@ -85,16 +88,51 @@ def get_rpath_direct(self, file_name: Path) -> str:
...
+            case "patchelf" | "lief-patchelf":
+                return _Patchelf(variant=patcher, platform=platform)
+            case "none":
+                return _NonePatcher(platform=platform)
+            case _:
+                msg = f"Unknown patcher {patcher!r}"
+                raise ValueError(msg)
+
+
+class _NonePatcher(ElfPatcher):
```
lkollar, line 103: I think `get_rpath` will raise `NotImplementedError` which is inconsistent with the error handling below. We should also cover that.

lkollar (commented):
Looks good, just a few minor concerns.

src/auditwheel/main_repair.py
mayeut: IMHO, all patchers should be valid on all platforms. The fact that `lief-patchelf` is only tested on a subset of platforms is because lief only provides those ones pre-built, not because they shouldn't work.

src/auditwheel/patcher.py
mayeut: The fact that `get_rpath` raises `NotImplementedError` is covered by unit tests.
The fact that it's not called when no patching occurs is covered by integration tests.

If this becomes a problem, we should move the `get_rpath_direct` implementation to `elftools` (as is already the case in `lddtree`) in the parent class.

lkollar (approved):

Issue #716, opened 2026-08-05
repair --strip strips after patchelf and can produce wheels that fail to import (ELF load command address/offset not page-aligned)


Fedr:
`repair --strip` strips the grafted libraries **after** patching them with patchelf (`strip_symbols()` runs last in `repair_wheel()`, after `copylib()` has already changed sonames and rpaths: https://github.com/pypa/auditwheel/blob/main/src/auditwheel/repair.py#L127-L130). Running strip on a patchelf-modified ELF is a long-standing corruption scenario — see NixOS/patchelf#145, NixOS/patchelf#371, NixOS/patchelf#507 — and when it triggers, the repaired wheel imports fine nowhere:

```
ImportError: libTKernel-fb973b98.so.7.9.1: ELF load command address/offset not page-aligned
```

## What happened

Building MeshLib manylinux wheels with `auditwheel repair --plat manylinux_2_28_{x86_64,aarch64} --strip`:

- `auditwheel repair` itself succeeds and logs `Stripping symbols from …` for every grafted lib.
- The resulting wheel fails to import with the error above on every distro we test (Rocky 8, Debian 11, Ubuntu 22.04/25.10, Fedora 37/39/42), on both x86_64 and aarch64.
- The corrupted library (`libTKernel` from OpenCASCADE) was **already stripped** before repair — `strip -s` rewrites the file layout even when it has no symbols to remove, so pre-stripping the inputs does not make `--strip` safe.

Public CI evidence (logs available):

- Failing run with `--strip`: https://github.com/MeshInspector/MeshLib/actions/runs/31016844725 — build jobs green, all 14 install/test jobs fail at import.
- Control run, identical except the libs are stripped *before* `auditwheel repair` (no `--strip`): https://github.com/MeshInspector/MeshLib/actions/runs/31020073470 — fully green, so stripping itself is fine; only the strip-after-patchelf order breaks.

## Environment

- auditwheel 6.7.0, patchelf 0.17.2.4 (both from PyPI), Python 3.12
- GNU strip from binutils 2.30 (Rocky Linux 8 container)
- Whether corruption triggers appears to depend on how much patchelf had to relocate; libs that receive a mangled soname plus an appended rpath (i.e. everything `copylib()` grafts) are rewritten enough to be vulnerable.

## Suggested fix

[...]

PR #718, opened 2026-08-08
test: always strip by default

mayeut:
check if it would have uncovered #716


mayeut:
it would have uncovered #716 and #719

PR #732, opened 2026-08-29
Reject wheels that require executable stacks

### LINKED ISSUE #634: Auditwheel shall check & reject shared objects with executable stack
At least Debian 13 has some hardening regarding SO built with executable stack which leads to error like the following
```
ImportError: libopenblas_armv6p-r0-eb0c233c.3.8.so: cannot enable executable stack as shared object requires: Invalid argument
```

Auditwheel shall check & reject shared objects with executable stack.
An option could also be added to clear the executable stack bit on grafted libraries, similar to what `execstack -c` [does](https://linux.die.net/man/8/execstack)

CAOShurong:
Fixes #634.

This checks each ELF `PT_GNU_STACK` program header for the execute flag. Wheels that request an executable stack are classified as Linux-only by `show`, reported in both text and JSON output, and explicitly rejected by `repair` for manylinux policies.

During repair, the check also covers external libraries that would actually be grafted for each candidate policy. The policy-specific evaluation rejects an unsafe target without over-rejecting another target that whitelists the same library and therefore does not copy it into the wheel.

The integration flow uses the real armv7l OpenBLAS case: repair must fail while the grafted library requests an executable stack and succeed after that flag is cleared.

Validation (exact head `ac90cb6deff8d525897c939562c82540ffd8a93e`):

- `pytest -q tests/unit/test_wheel_abi.py tests/integration/test_bundled_wheels.py::test_repair_rejects_executable_stack` — 13 passed
- `pre-commit run --files src/auditwheel/wheel_abi.py tests/unit/test_wheel_abi.py` — all hooks passed, including Ruff and mypy
- `pytest -q tests/unit/test_main_show.py -k executable_stack` — 2 passed, covering the JSON flag and text warning path
- `git diff --check`

The ARM container matrix is not available locally and is left to upstream CI.

AI assistance disclosure: OpenAI Codex assisted with the implementation and tests; I reviewed the resulting diff and validation output before pushing it.


src/auditwheel/show-schema.json
mayeut:
```suggestion
          "const": 2,
```
The version shall be updated to reflect the addition of the `executable_stack` member

mayeut (dismissed):
Thanks for the PR
Can you update one of the integration tests using `build_numpy` to check for failure before patching openblas ? (requires moving openblas patch outside `build_numpy`)

src/auditwheel/wheel_abi.py
```diff
@@ -143,6 +147,8 @@ def get_wheel_elfdata(
 
         platform_wheel = False
         for fn, elf in elf_file_filter(ctx.iter_files()):
+            if elf_has_executable_stack(elf):
+                executable_stack.append(fn)
```
mayeut, line 151: this only applies to elf files already present in the wheels and thus misses grafted libs which explains the failing tests.

mayeut (changes requested):

auvipy (commented):
looks good

sylvesterkaczmarek (commented):
elf_has_executable_stack() returns False when PT_GNU_STACK is missing. A missing marker is not the same as an explicit non-executable GNU_STACK; legacy/unknown ELF can still be treated as needing an executable stack. For a reject-on-exec-stack check, can the no-GNU_STACK case be handled as unknown/unsafe rather than safe, with a fixture for it?

